PR #86, opened 2018-04-23
[WIP] Add API reference

rijkvanzanten:
_To be hosted on GH Pages_

This PR is the start of migrating the API reference into a VuePress based static site that is going to be hosted on GH Pages (in this repo).

Deploying the docs should (and will) be automated using a CI, but that's not a worry for right now.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -17,7 +17,7 @@ A special character ( `_` ) can be used to target the default database, whicheve
 
 A few examples of api requests:
 
-*   `/api/_/collections/` (uses default config file `config.php`)
+*   `/api/_/collections` (uses default config file `config.php`)
 *   `/api/prod/items/projects` (uses prod config file `config.prod.php`)
 
 ::: tip
@@ -149,7 +149,7 @@ The API will send an email to the requested user’s email containing a link wit
 **@TODO**: Is this correct? The reset token could also use the JWT spec to keep it consistent. The payload should contain the email address and expiration date.
 
 ```http
-POST /auth/password/request/
+POST /auth/password/request
 ```
 
 #### Body
@@ -301,7 +301,7 @@ Items are essentially individual database records which each contain one or more
 Creates one or more items in a given collection
 
 ```http
-POST /items/[collection-name]/
+POST /items/[collection-name]
 ```
 
 #### Body
@@ -386,7 +386,7 @@ GET /items/[collection-name]/[pk],[pk],[pk]
 Get an array of items from a given collection
 
 ```http
-GET /items/[collection-name]/
+GET /items/[collection-name]
 ```
 
 #### Query Parameters
@@ -896,7 +896,7 @@ The path to the last page the user was on in the Directus App
 Hashes the submitted string using the chosen algorithm
 
 ```http
-POST /utils/hash/
+POST /utils/hash
 ```
 
 #### Body
@@ -949,7 +949,7 @@ The hashing algorithm to use and the string to hash
 Gets a random alphanumeric string from the API
 
 ```http
-GET /utils/random/
+GET /utils/random
 ```
 
 | Name   | Default | Description                |
--- a/docs/README.md
+++ b/docs/README.md
@@ -860,7 +860,7 @@ or
 Set the time and last Directus App page accessed by the user. Last Access is used to determine if the user is still logged into the Directus app, and Last Page is used to avoid editing conflicts between multiple users.
 
 ```http
-POST /users/tracking/page
+PATCH /users/[pk]/tracking/page
 ```
 
 #### Body
--- a/docs/README.md
+++ b/docs/README.md
@@ -614,6 +614,10 @@ These system endpoints still follow the same spec as a “regular” `/items/[co
 
 `/files` is used for creating or updating a file requires the API to accept a special field allowing for the base64 file data. Beyond that, it accepts POST requests with the multipart-formdata enctype, to allow for easier uploading of file(s).
 
+### Folders
+
+`/folders` is used for creating, updating, or deleting a virtual folder.
+
 ### Permissions
 
 `/permissions` does not have any additional processing, it is simply an alias for the blocked `/items/directus_permissions`
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,7 +7,8 @@ sidebar: auto
 ## Introduction
 
 ### Versioning
-The Directus API uses SemVer for version labeling within the repo and for files which mention a specific version (eg: `package.json`). The API will *not* include the version in the URL because the API is "versionless". Being versionless means that we will not remove or change API behavior, only *adding* new features and enhancements – therefore, no breaking changes will ever be introduced.
+
+The Directus API uses SemVer for version labeling within the repo and for files which mention a specific version (eg: `package.json`). The API will _not_ include the version in the URL because the API is "versionless". Being versionless means that we will not remove or change API behavior, only _adding_ new features and enhancements – therefore, no breaking changes will ever be introduced.
 
 ### Environments
 
@@ -15,8 +16,9 @@ All endpoints are prefixed with the database name (as defined in the config file
 A special character ( `_` ) can be used to target the default database, whichever that is.
 
 A few examples of api requests:
-* `/api/_/collections/` (uses default config file `config.php`)
-* `/api/prod/items/projects` (uses prod config file `config.prod.php`)
+
+*   `/api/_/collections/` (uses default config file `config.php`)
+*   `/api/prod/items/projects` (uses prod config file `config.prod.php`)
 
 ::: tip
 The name in the API URL is not the name of the database itself.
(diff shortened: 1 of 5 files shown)


docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+| Code             | Description                                        |
+|------------------|----------------------------------------------------|
+| 200 OK           | `data`: Retrieved item<br>`meta`: Depends on requested metadata  |
+| 404 Not Found    | `message`: Collection doesn’t exist, or item doesn't exist within collection    |
+
+#### Examples
+
+* Return the project item with an ID of `1`
+  ```bash
+  curl -g https://api.directus.io/_/items/projects/1
```
rijkvanzanten: The `-g` flag in `curl` is only needed when you want to use brackets in query parameters like filter **[** name **]** **[** eq **]** =rijk

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+  ```
+
+
+
+
+### Get Items
+Gets items from within a given collection
```
rijkvanzanten: From _within_?

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+| limit         | 200       | The number of items to request                   |
+| offset        | 0         | How many items to skip before fetching results   |
+| sort          | id        | CSV of fields to sort by [Learn More](#sorting)  |
+| fields        | *         | CSV of fields to include in response [Learn More](#fields)  |
+| filter[field] |           | Filter items using operators [Learn More](#filtering)  |
+| meta          |           | CSV of metadata fields to include [Learn More](#metadata)  |
+| status        | Published | CSV of statuses [Learn More](#status)  |
+| lang          | *         | Include translation(s) [Learn More](#language)  |
+| q             |           | Search string [Learn More](#search-query)  |
+| id            |           | CSV of primary keys to fetch  |
```
rijkvanzanten: Is this an actual existing parameter? Is it called `id`? We use the term primary key everywhere now, so this feels very outdated. If you want to fetch items by primary key, use `/items/projects/<pk>,<pk>,<pk>` instead

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+Update a single item from within a given collection
+
+```http
+PATCH /items/[collection-name]/[primary-key-value]
+PUT /items/[collection-name]/[primary-key-value]
+```
+
+::: warning PATCH VS PUT
+* **PATCH** partially updates the item with the provided data, any missing data is ignored
+* **PUT** fully updates the item and any missing data will fallback to its default value
```
rijkvanzanten: ~fully updates~ => overwrites

I think the term overwrites expresses the actual functionality better

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+:::
+
+
+#### Body
+A single item to be updated. Field keys must match the collection's column names
+
+#### Possible Results
```
rijkvanzanten: I'm not sure if the term `Possible Results` is the best.. For all we know, the API might return a 500 or other due to unexpected reasons. These results are the _most likely_ results..

rijkvanzanten: Also, 403 is a possible result for _all_ endpoints, seeing that the "public"-ness of an endpoint is set by the permissions, any endpoint can potentially return a 403 both on initial request as on the return value. 

F.e. if your user has the right to update, the update itself will go through, but it will still return 403..

rijkvanzanten: Which makes me think, in case you **don't have read access** but **do have update** access, we should probably return a response in the 2xx-range

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+| 200 OK                   | `data`: The updated item, including default fields added by Directus  |
+| 400 Bad Request          | `message`: Syntax error in provided JSON    |
+| 404 Not Found            | `message`: Collection doesn’t exist    |
+| 422 Unprocessable Entity | `message`: Column doesn’t exist in collection    |
+
+#### Examples
+
+* Return the project item with an ID of `1`
+  ```bash
+  curl -g https://api.directus.io/_/items/projects/1
```
rijkvanzanten: -g

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+::: warning PATCH VS PUT
+* **PATCH** partially updates the item with the provided data, any missing data is ignored
+* **PUT** fully updates the item and any missing data will fallback to its default value
+:::
+
+#### Body
+An array of items to be updated. Field keys must match the collection's column names. Alternatively, you can use the following syntax to overwrite the values of multiple items at once (Batch Update).
+
+```json
+{
```
rijkvanzanten: This is not valid JSON

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+  ```bash
+  curl -g https://api.directus.io/_/items/projects/1
+  ```
+
+
+
+
+### Update Items
+Update multiple items within a given collection
+@TODO WHY DOES THIS NOT USE A CSV OF IDS WITH UPDATE ITEM?
```
rijkvanzanten: This is an "older" spec.. Let me confirm if this is actually working in the API:
https://github.com/directus/api/issues/91

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+
+### Delete Items
+Deletes multiple items from a specific collection
+
+```http
+DELETE /items/[collection-name]/
+```
+
+#### Body
```
rijkvanzanten: This should be `DELETE /items/[collection-name]/[pk],[pk]` ref https://github.com/directus/api/issues/91

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+An array of IDs of records to be deleted @TODO SHOULD THIS BE PKs?
+
+#### Possible Results
+| Code                     | Description                                        |
+|--------------------------|----------------------------------------------------|
+| 202 No Content @TODO 204?                  | Items were deleted successfully  |
```
rijkvanzanten: Yes!

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+| 422 Unprocessable Entity | `message`: Items don't exist in collection    |
+
+
+
+
+
+
+## System
+
+All system tables (`directus_*`) are blocked from being used through the regular `/items` endpoint to prevent security leaks or because they require additional processing before sending to the end user. This means that any requests to `/items/directus_*` will always return `403 Forbidden`.
```
rijkvanzanten: It might return a 404.. we need to check that

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+These system endpoints still follow the same spec as a “regular” `/items/[collection-name]` endpoint but require the additional processing outlined below:
+
+### Columns
+`/columns` is used for creating, updating, or deleting columns through the API requires the API to modify the database schema directly. @TODO
+
+### Files
+`/files` is used for creating or updating a file requires the API to accept a special field allowing for the base64 file data. Beyond that, it must accept POST requests with the multipart-formdata enctype, to allow for easier uploading of file(s), and must accept uploading by chunk, to allow for larger filesizes. @TODO
+
+### Permissions
```
rijkvanzanten: Nothing really 😄 Just for consistencies sake with the other `directus_*` collections, it has it's own endpoint

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+| Code             | Description                                        |
+|------------------|----------------------------------------------------|
+| 200 OK           | `data`: Retrieved user<br>`meta`: Depends on requested metadata  |
+| 404 Not Found    | `message`: Item doesn't exist within collection    |
+
+#### Examples
+
+* Return the user with an ID of `1`
+  ```bash
+  curl -g https://api.directus.io/_/users/1
```
rijkvanzanten: -g

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+```http
+GET /users
+```
+
+#### Query Parameters
+| Name          | Default   | Description                                      |
+|---------------|-----------|--------------------------------------------------|
+| meta          |           | CSV of metadata fields to include [Learn More](#metadata)  |
+@TODO DOES THIS REALLY NOT ALLOW FOR FILTERING OFFSET ETC?
```
rijkvanzanten: It does. It should accept exactly the same parameters as GET ITEMS, just like all other GET system endpoints

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+#### Possible Results
+| Code             | Description                                        |
+|------------------|----------------------------------------------------|
+| 200 OK           | `data`: Array of Directus users<br>`meta`: Depends on requested metadata  |
```
rijkvanzanten: I guess a 400 Bad Request is possible too if you mess up the query parameters. Don't think that's implemented though.

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+| 422 Unprocessable Entity | `message`: Column doesn’t exist in collection    |
+
+
+
+
+### Delete User
+Deletes a user from this instance
+
+```http
+DELETE /users/[primary-key-value]
```
rijkvanzanten: Or multiple with multiple primary keys, ref https://github.com/directus/api/issues/91

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+```json
+{
+  "email": "rijk@example.com"
+}
+```
+or
+```
+{
+  "email": [
```
rijkvanzanten: Maybe we should use `rijk@rngr.org` instead of `rijk@example.com`? 😄

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+```json
+{
+  "last_page": "/tables/projects"
+}
+```
+
+#### Possible Results
+| Code                     | Description                                        |
+|--------------------------|----------------------------------------------------|
+| 200 OK                   | Emails successfully sent @TODO IS THIS USED?  |
```
rijkvanzanten: Email send on tracking user page?! I don't think so!

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+
+
+
+## Extensions
+
+Directus can easily be extended through the addition of several types of extensions. Extensions are important pieces of the Directus App that live in the decoupled Directus API. These include Interfaces, Listing Views, and Pages. These three different types of extensions  live in their own directory and may have their own endpoints.
+
+### Get Interfaces, List Views, Pages
+These endpoints read the API's file system for directory names and return an array of extension names.
```
rijkvanzanten: AFAIK, it also returns the `meta.json` file contents directly in this request

docs/README.md
```diff
@@ -33,42 +37,736 @@ All output will adhere to the same general JSON structure:
...
+
+```http
+GET /interfaces
+GET /listviews
+GET /extensions
+```
+@TODO extensions -> pages
```
rijkvanzanten: I think this is implemented already

docs/README.md
benhaynes: Yeah, it fetches the items that are within a collection...

Get's 1,2,3 from within `projects`

Does that not sound right?

PR #167, opened 2018-05-30
Add missing field interface values

honzabilek4:
This was causing an uncaught PDO exceptions on `install:database` command.


WellingGuzman:
That looks good @honzabilek4! Thanks!

PR #173, opened 2018-06-06
Add check for existence of primary key field on collection creation / item retrieval 

### LINKED ISSUE #172: Collections can be created without a primary key, which makes the collection unusable.
**Describe the bug**
A user is able to create a collection where none of the fields are set as the primary key. This causes issues when the user attempts to do anything with the collection, such as retrieve the fields for the collection, because it throws an error stating that the table has no primary key.

**To Reproduce**
Steps to reproduce the behavior

1. send a post request to `/collections?access_token=admin_token` with the body:

```json
{
    "collection":"test",
    "single":false,
    "hidden": false,
    "note":"hello world",
    "fields":[
        {
            "type":"integer",
            "field":"field1",
            "interface":"check"
        }
    ]
}
```

2. send a get request to `/items/test1?access_token=admin_token`

3. The response should contain `"message": "This row object does not have a primary key column set."`

**Expected behavior**
If the primary key is required, then I would expect to get a 400 when I attempt to create the collection.

danielfrentz:
Added some validation to ensure that at least 1 field has primary_key set to true. If this is not the case then it throws a BadRequestException with the error message "Collection does not have a primary key."

Closes #172

### DIFF
--- a/src/core/Directus/Services/TablesService.php
+++ b/src/core/Directus/Services/TablesService.php
@@ -217,6 +217,7 @@ class TablesService extends AbstractService
      * @throws InvalidRequestException
      * @throws CollectionAlreadyExistsException
      * @throws UnauthorizedException
+     * @throws BadRequestException
      */
     public function createTable($name, array $data = [], array $params = [])
     {
--- a/src/core/Directus/Services/TablesService.php
+++ b/src/core/Directus/Services/TablesService.php
@@ -743,7 +743,13 @@ class TablesService extends AbstractService
         return SchemaService::getCollection($tableName);
     }
 
-    public function hasPrimaryField($fields){
+    /**
+     * @param $fields
+     * @return bool
+     *
+     * Checks that at least one of the fields has primary_key set to true.
+     */
+    public function hasPrimaryField(array $fields){
         $result = false;
 
         foreach($fields as $field){
@@ -755,7 +761,14 @@ class TablesService extends AbstractService
         return $result;
     }
 
-    public function hasUniquePrimaryKey($fields){
+    /**
+     * @param $fields
+     * @return bool
+     *
+     * Checks that a maximum of 1 field has the primary_key field set to true. This will succeed if there are 0
+     * or 1 fields set as the primary key.
+     */
+    public function hasUniquePrimaryKey(array $fields){
         $primaryKeyCount = 0;
         foreach($fields as $field){
             if($field['primary_key'] === true){
--- a/src/core/Directus/Services/TablesService.php
+++ b/src/core/Directus/Services/TablesService.php
@@ -258,6 +258,9 @@ class TablesService extends AbstractService
         if(!$this->hasPrimaryField($data['fields'])){
             throw new BadRequestException("Collection does not have a primary key.");
         }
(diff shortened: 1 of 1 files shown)


src/core/Directus/Services/TablesService.php
```diff
@@ -735,6 +740,19 @@ public function getTableObject($tableName)
         return SchemaService::getCollection($tableName);
     }
 
+    public function hasPrimaryField($fields){
+        $result = false;
+
+        foreach($fields as $field){
+            if($field[primary_key] == true){
```
WellingGuzman: @danielfrentz it looks great, but you have to make sure the array key is a string instead of a constant.

`$field['primary_key']` instead of `$field[primary_key]`

rijkvanzanten: And maybe use a strict equal check? `===` vs `==` 🙂

src/core/Directus/Services/TablesService.php
```diff
@@ -735,6 +743,29 @@ public function getTableObject($tableName)
...
+    }
+
+    public function hasUniquePrimaryKey($fields){
+        $primaryKeyCount = 0;
+        foreach($fields as $field){
+            if($field['primary_key'] === true){
+                $primaryKeyCount++;
+            }
+        }
+        return $primaryKeyCount <= 1;
```
WellingGuzman: Hey @danielfrentz, this must be `$primaryKeyCount === 1` as this will fail when there's not primary key by returning `true`. When checking the condition `0 <= 1`.

src/core/Directus/Services/TablesService.php
```diff
@@ -735,6 +743,29 @@ public function getTableObject($tableName)
...
+        foreach($fields as $field){
+            if($field['primary_key'] === true){
+                $result = true;
+                break;
+            }
+        }
+        return $result;
+    }
+
+    public function hasUniquePrimaryKey($fields){
```
WellingGuzman: Adding the type of `$fields` in the function declaration parameters make it avoid us to add validation to the function.

Example: `public function hasUniquePrimaryKey(array $fields)`

src/core/Directus/Services/TablesService.php
danielfrentz: This may be too small of a point to argue about but heres my case for <= over ===.

Theres 2 separate checks. The first checks to see if the key exists at all, while this one checks that there is a maximum of 1. 
If it were changed to === then from a maintainability point of view, it would make life a little harder since if you remove the other check, or move it so that it happens afterwards,  then this one will trigger when theres only 1 primary key and give a bad error.

WellingGuzman: The intention of this is to tell whether there's only or just if there's not more than one?

These are two public function that does a separated validation and it shouldn't depend on a other function result indirectly, like whether or not the first failed or not.

Please add comments to the functions definition to make sure what really does, and it should be good!

benhaynes: Hey @danielfrentz – any update on this? We'd love to get it merged/closed! Thanks again!

WellingGuzman:
Thanks @danielfrentz!

PR #189, opened 2018-06-13
-Changed the error code of CollectionNotFound to 404.

danielfrentz:
-Added a check to ensure that there was at least 1 item found.
-Added a check that the results found in relationalTableGateway was
actually an array - fixes a warning which was thrown, which caused the 203

### DIFF
--- a/src/core/Directus/Database/Exception/CollectionNotFoundException.php
+++ b/src/core/Directus/Database/Exception/CollectionNotFoundException.php
@@ -6,7 +6,7 @@ use Directus\Exception\NotFoundException;
 
 class CollectionNotFoundException extends NotFoundException
 {
-    const ERROR_CODE = 200;
+    const ERROR_CODE = 404;
 
     public function __construct($name)
     {
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -1019,7 +1019,7 @@ class RelationalTableGateway extends BaseTableGateway
         $meta = ArrayUtils::get($params, 'meta', false);
         $id = ArrayUtils::get($params, 'id');
         $single = ArrayUtils::get($params, 'single');
-        if (!$single && $id && count($id) == 1) {
+        if (!$single && $id && is_array($id) && count($id) == 1) {
             $single = $params['single'] = true;
         }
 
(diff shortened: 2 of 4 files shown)


src/core/Directus/Database/Exception/CollectionNotFoundException.php
```diff
@@ -6,7 +6,7 @@
 
 class CollectionNotFoundException extends NotFoundException
 {
-    const ERROR_CODE = 200;
+    const ERROR_CODE = 404;
```
WellingGuzman: Nice catch @danielfrentz!

src/core/Directus/Database/TableGateway/RelationalTableGateway.php
```diff
@@ -1019,7 +1019,7 @@ protected function fetchData(array $params = [])
         $meta = ArrayUtils::get($params, 'meta', false);
         $id = ArrayUtils::get($params, 'id');
         $single = ArrayUtils::get($params, 'single');
-        if (!$single && $id && count($id) == 1) {
+        if (!$single && $id && is_array($id) && count($id) == 1) {
```
WellingGuzman: This one has been fixed already by https://github.com/directus/api/commit/bbf97f68659995bd6279f910c9501f4ba09f2f7c

src/core/Directus/Services/TablesService.php
```diff
@@ -102,7 +102,12 @@ public function findByIds($name, array $params = [])
 
         $tableGateway = $this->createTableGateway($this->collection);
 
-        return $tableGateway->getItemsByIds($name, $params);
+        $collection = $tableGateway->getItemsByIds($name, $params);
+        if(count($collection['data']) === 0){
```
WellingGuzman: Find by Ids get into the the "filtering" category, which means if there's 0 result, is a valid response.

src/endpoints/Collections.php
```diff
@@ -67,12 +68,17 @@ public function all(Request $request, Response $response)
...
+        $name = $request->getAttribute('name');
         $service = new TablesService($this->container);
         $responseData = $service->findByIds(
-            $request->getAttribute('name'),
+            $name,
             $request->getQueryParams()
         );
 
+        if(count($responseData['data']) === 0){
+            throw new CollectionNotFoundException($request->getAttribute($name));
```
WellingGuzman: `$request->getAttribute($name)` will not get the expected value, as if I request `/collections/users`, it will try to get an attribute named `users` `$request->getAttribute('users')`

I am glad you are taking your time to get this PRs, but the collections and fields endpoint are more complex and it needs set the tables service where it can be shared across the codebase.

If you remove these changes I will merge it and we can discuss more about your ideas to solve this issue.

WellingGuzman:
Thanks @danielfrentz!

src/core/Directus/Database/Exception/CollectionNotFoundException.php
WellingGuzman: Actually @danielfrentz I am so sorry I let you waste your time.

This is the error code, not the http status code.

PR #241, opened 2018-07-10
[WIP] Add calendar interface

jsguru-git:
I added calendar interface into extensions/core/interfaces.

### DIFF
--- a/extensions/core/interfaces/calendar/interface.vue
+++ b/extensions/core/interfaces/calendar/interface.vue
@@ -9,47 +9,49 @@
 </template>
 
 <script>
-import flatPickr from 'vue-flatpickr-component';
+import flatPickr from "vue-flatpickr-component";
 // Need to add base css for flatpickr
-import 'flatpickr/dist/flatpickr.css';
-import './calendar.css';
+import "flatpickr/dist/flatpickr.css";
+import "./calendar.css";
 // import 'flatpickr/dist/themes/material_red.css';
-import mixin from '../../../mixins/interface';
+import mixin from "../../../mixins/interface";
 
 export default {
-	name: 'interface-calendar',
-	computed: {
-		// dateFormat() {
-		// 	if (this.options.formatting == "" || this.options.formatting == null) {
-		// 		return "Y-m-d";
-		// 	}
-		// 		return this.options.formatting;
-		// },
-		configs() {
-			return {
-				inline: true,
-				minDate: this.options.min,
-				maxDate: this.options.max,
-				dateFormat: "Y-m-d",
-				// dateFormat: this.dateFormat,
-				// mode: "range"
-			}
-		},
-	},
-	mixins: [mixin],
-	components: {
-    flatPickr
+  name: "interface-calendar",
+  computed: {
+    // dateFormat() {
+    // 	if (this.options.formatting == "" || this.options.formatting == null) {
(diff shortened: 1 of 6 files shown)


extensions/core/interfaces/calendar/interface.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+	<div v-bind:class="{ inactive: readonly }">
+		<flat-pickr class="form-control"
+                :config="configs"
+                v-model="value"
```
rijkvanzanten: This throws an error. `value` is being passed in as a prop from the parent, and is being updated by `$emit`ing the `input` event with a value.

extensions/core/interfaces/calendar/meta.json
```diff
@@ -0,0 +1,39 @@
...
+      "name": "$t:min",
+      "comment": "$t:min_comment",
+      "interface": "date"
+    },
+    "max": {
+      "name": "$t:max",
+      "comment": "$t:max_comment",
+      "interface": "date"
+    },
+    "formatting": {
```
rijkvanzanten: This isn't used yet, correct?

extensions/core/interfaces/calendar/readonly.vue
```diff
@@ -0,0 +1,11 @@
+<template>
+  <div class="no-wrap">{{ value }}</div>
```
rijkvanzanten: I think it would be nice if there's an option to set the value to "relative" mode, just like in the `date` interface 😄

extensions/core/interfaces/calendar/interface.vue
```diff
@@ -0,0 +1,81 @@
...
+		},
+	},
+	mixins: [mixin],
+	components: {
+    flatPickr
+  },
+}
+</script>
+
+<style lang="scss">
```
rijkvanzanten: The style [needs to be `scoped`](https://vue-loader.vuejs.org/guide/scoped-css.html), otherwise (generic) classes can and will cause problems later on. For example, we don't want this `max-width` on all components that use a `.inactive` class.

rijkvanzanten:
@benhaynes We just decided to leave out the `localized` option seeing we already have that format option. One will break the other and is therefore very confusing.

benhaynes:
Perfect! So this one is ready to merge? Let's do it :)

rijkvanzanten:
@benhaynes I've asked the creator of the timeago component if it's possible to have it return _Today_, _Tomorrow_, etc. https://github.com/egoist/vue-timeago/issues/83. If not, we have to roll our own custom function that returns those things on the right days.

rijkvanzanten:
A more immediate fix would be a conditional that checks “if date === today return ‘today’”

benhaynes:
Works perfectly!

We will have to fix the relative at some point... (if the main repo isn't fixed)I'm getting `In 1 day` when it's TWO days away. And I'm getting hours for yesterday, today, and tomorrow. But other than that this is good!

rijkvanzanten:
> I'm getting In 1 day when it's TWO days away

I feel like it's comparing your local timezone (-5h) with the saved timezone (UTC) which causes the problems in non-working times

benhaynes:
Yeah, must just use `00:00:00` ... so however many hours you are to that day at midnight, it turns to hours. Hopefully we/they can fix!

PR #242, opened 2018-07-12
Added Star-Rating Interface

amigo1205:


extensions/.htmlnanorc
```diff
@@ -0,0 +1,3 @@
+{
+  minifySvg: false
```
rijkvanzanten: Is this necessary for this interface to work properly? I'd like to keep this minification on is possible

extensions/core/interfaces/star-rating/star-rating.vue
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        this.$set(this.fillLevel, i, Math.round(level));
+      }
+    }
+  },
+  computed: {
+    activeColor() {
+      if (this.color == "blue") {
+        return "#039be5";
```
rijkvanzanten: These should use the global available variables. F.e. `var(--blue)` instead of `#039be5`

extensions/core/interfaces/star-rating/star.vue
```diff
@@ -0,0 +1,178 @@
...
+    },
+    glowColor: {
+      type: String,
+      required: false
+    }
+  },
+  created() {
+    this.starPoints = this.points.length ? this.points : this.starPoints;
+    
+    console.log(this.starPoints);
```
rijkvanzanten: We try to prevent console.log statements in production code

extensions/core/interfaces/star-rating/star.vue
```diff
@@ -0,0 +1,178 @@
...
+      grad: "",
+      glowId: ""
+    };
+  }
+};
+</script>
+
+<style scoped>
+.vue-star-rating-star {
+  overflow: visible !important;
```
rijkvanzanten: Does this need `!important`?

extensions/.htmlnanorc
amigo1205: Parcel bundler doesn't work for SVG (with Vue Directives).
To use SVG component, I had to use this configuration.

benhaynes:
Just reiterating our Slack feedback from a few days ago here:

• We need an option for: `Allow Decimals` (`false` by default)
• Instead of a blue/yellow toggle, let's use the `color` interface with Blue, Yellow, and Darkest-Gray as the three options
• The value is not being saved
• There should be a value (smaller stars) in the "Listing" view
• You can remove the "Disabled State" option... this is controlled by the global `readonly`  option
• When I set max rating to `10` it no longer works
• When I change the value in the text input it doesn't update the stars

Generally, this seems over-engineered. Users won't click 4.2 stars... if they need that precision they will enter it in the input. The star icons should just use the three Material Design icons (empty, half, full) for display... but when clicking you should only enter `full` stars to keep things simple.

extensions/.htmlnanorc
rijkvanzanten: Ref https://github.com/parcel-bundler/parcel/issues/1124

rijkvanzanten:
This is unfortunately too overkill of a solution for the problem at hand. The thinking was right and the code does as advertised, but we're looking for something simpler for the core rating interface.

PR #243, opened 2018-07-13
add code editor interface. At this point, click the paste icon to att…

jsguru-git:
…ach a placeholder. This feature prevent syncing code editing with VALUE and LISTING.


rijkvanzanten:
Looking good so far! Not too sure about the `readonly.vue` one though. Right now, it'll just render the whole value on listings / other places in the app, but seeing this is a code editor, that's probably getting out of hand really quickly. We could just render a static icon indicating that it's a code snippet, just like I did for the JSON interface. Thoughts @benhaynes @jsguru-git?

extensions/core/interfaces/code/code.css
```diff
@@ -0,0 +1,14 @@
+.CodeMirror {
+  /*color: var(--gray);*/
+  font-weight: inherit;
+  /*height: auto;*/
```
rijkvanzanten: Other nitpicky thing: I'd like to remove code instead of comment it out. We have git to rely on if we ever want to revert back to code. Removing deprecated code keeps it nice and clean.

extensions/core/interfaces/code/interface.vue
```diff
@@ -0,0 +1,161 @@
+<template>
+  <div class="interface-code">
+    <codemirror ref="cm"
```
rijkvanzanten: I personally prefer to move the first attribute to a new line and indent the following ones on the same level, that way we prevent a big indent "gap":

```vue
<!-- Before -->
<codemirror ref="cm"
            :value="options.placeholder"
            :options="cmOptions"
            @ready="onCmReady"
            @focus="onCmFocus"
            @input="onCmCodeChange"
></codemirror>

<!-- After -->
<codemirror
  ref="cm"
  :value="options.placeholder"
  :options="cmOptions"
  @ready="onCmReady"
  @focus="onCmFocus"
  @input="onCmCodeChange">
</codemirror>
```

extensions/core/interfaces/code/interface.vue
```diff
@@ -0,0 +1,161 @@
...
+      <button class="btn" @click="$emit('input', code)">
+        <i class="material-icons">assignment_turned_in</i>
+      </button>
+    </div>
+  </div>
+</template>
+
+<script>
+
+// require component
```
rijkvanzanten: I'm always in favor of descriptive comments explaining the _why_, but comments saying "require css" followed by an `import x from "y.css"` aren't really adding any value, but do add visual clutter

extensions/core/interfaces/code/interface.vue
```diff
@@ -0,0 +1,161 @@
...
+
+import './code.css';
+import mixin from '../../../mixins/interface';
+
+export default {
+  name: "interface-code",
+  mixins: [mixin],
+  data () {
+    return {
+      code: "",
```
rijkvanzanten: There's a `npm run lint` and a matching `npm run fix` command to make the code lint free according to the same specs as the rest of the app 🙂

extensions/core/interfaces/code/interface.vue
```diff
@@ -0,0 +1,161 @@
...
+        extraKeys: { "Ctrl": "autocomplete" }
+      }
+    },
+    codemirror() {
+      return this.$refs.cm.codemirror;
+    }
+  },
+  methods: {
+    onCmReady(cm) {
+      console.log('the editor is readied!', cm)
```
rijkvanzanten: We should avoid console.log statements in production code. Let's remove these methods unless they're going to be used for something.

rijkvanzanten:
The name `code` seems to cause errors in Vue:

```
[Vue warn]: Do not use built-in or reserved HTML elements as component id: code
```

rijkvanzanten:
Can we figure out a way to load the languages async @jsguru-git? Would be nice to support all available languages in codemirror.

rijkvanzanten:
@benhaynes We could support multiple control modes too! Right now it's hardcoded to sublime, but I'm very much in favor of having the option to use Vim controls 😎

rijkvanzanten:
Parcel (which is used for the extensions buildchain) has [support for codesplitting](https://parceljs.org/code_splitting.html). We might be able to load the required extra languages / options with `import()`? 😄

benhaynes:
@jsguru-git – here is my feedback:

* min/max-height should have default values. Let's use: 300/600
* The height options are not working right now. With the defaults, the editor STARTS at MIN-HEIGHT tall when blank, and can expand up to MAX-HEIGHT (after that it scrolls)
* Let's use title case for language options: `Javascript`, `Vue`, `PHP`, `JSON`... (not`JAVASCRIPT`)
* When hiding line numbers, Is it possible to add some padding to the editor? (same as `textarea` interface) 
* Paste in placeholder only works ONCE. If you paste, then change value and paste again it doesn't do anything. It should always work.
* How difficult is it to have a Option (toggle) for minify code? When off, all new lines and whitespace is preserved (for legibility)... when ON all of that is removed.

PR #245, opened 2018-07-16
Add collections interface

benhaynes:
Pulls from the store, should be working well

### DIFF
--- a/extensions/core/interfaces/collections/interface.vue
+++ b/extensions/core/interfaces/collections/interface.vue
@@ -1,0 +1,46 @@
+<template>
+  <v-select
+    :value="value"
+    :disabled="readonly"
+    :id="name"
+    :options="choices"
+    :placeholder="options.placeholder"
+    @input="$emit('input', $event)" />
+</template>
+
+<script>
+import mixin from "../../../mixins/interface";
+
+export default {
+  mixins: [mixin],
+  computed: {
+    choices() {
+      const collections = this.$store.state.collections || {};
+      let choices = {};
+
+      for (var key in collections) {
+        if (collections.hasOwnProperty(key) && collections[key].collection) {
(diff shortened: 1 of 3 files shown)


extensions/core/interfaces/collections/interface.vue
```diff
@@ -0,0 +1,46 @@
...
+export default {
+  mixins: [mixin],
+  computed: {
+    choices() {
+      const collections = this.$store.state.collections || {};
+      let choices = {};
+
+      for (var key in collections) {
+        if (collections.hasOwnProperty(key) && collections[key].collection) {
+          if (this.options.include_system || collections[key].collection.substring(0, 9) != 'directus_'){
```
rijkvanzanten: `collections[key].collection.substring(0, 9) != 'directus_'` could also be written as

`collections[key].collection.startsWith("directus_") === false`, which is IMO a little more readable

rijkvanzanten: Also, `'directus_'` => `"directus_"` 🤓

extensions/core/interfaces/collections/interface.vue
```diff
@@ -0,0 +1,46 @@
...
+            choices[key] = this.$helpers.formatTitle(collections[key].collection);
+          }
+        }
+      }
+
+      if (!choices) return {};
```
rijkvanzanten: Choices is by default an empty object, so this if statement is unnecessary

extensions/core/interfaces/collections/interface.vue
```diff
@@ -0,0 +1,46 @@
...
+
+      // if (typeof this.options.choices === "string") {
```
rijkvanzanten: Prefer using git for deprecated code over comment blocks

PR #257, opened 2018-07-19
fixes for the feature icons -> icon

VincentKempers:
- Changed the name from icons to icon
- Fixed the read-only.
- Deleted the tooltip, didn't add anything to the icon interface.
- add the read-only to the search results


rijkvanzanten:
@VincentKempers seems like the git history got messed up again 😳

PR #288, opened 2018-08-04
Map Interface

theharshin:
### Features Added:
- Button to center on the current location of the user. 
- Clicking on map moves a marker to new lat/lng & saves value.
- 'Clear' button clears the value & removes marker.
- Display currently marked lat/lng at bottom of map.
- 2 styling added (Colored, Black & White)
- Configurable [Width, Height, Zoom, MaxZoom, Default lat/lng]


rijkvanzanten:
The display.vue component is being used to render the value of the field in other places in the system, like the tabular / tiles view. While it's very cool that the main code is being re-used, rendering the full map on the layout views for every single row on the screen is very overkill.

Instead, let's render the `location_on` icon with a `v-tooltip` that shows the full value.

extensions/core/interfaces/map/display.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <div class="readonly-map">
+    <div class="interface-map-wrap">
+      <div class="map-display" id="directusMapDisplay" :style="{width:options.width+'px',height:options.height+'px'}">
```
rijkvanzanten: `id="directusMapDisplay"`

Keep in mind that this interface can be rendered multiple times on the page, which will cause duplicate ID issues! Field names are unique and can be accessed through `this.name`

extensions/core/interfaces/map/display.vue
```diff
@@ -0,0 +1,69 @@
...
+  name: "readonly-map",
+  mixins: [mapMixin],
+  data() {
+    return {
+      mapPlaceholder: "directusMapDisplay",
+      mode: "display"
+    };
+  },
+  watch: {
+    //? Do we need to re-render map when "value" changes from interface debugger?
```
rijkvanzanten: Yes. The value can be set by other interfaces as well, not just the interface debugger!

extensions/core/interfaces/map/display.vue
```diff
@@ -0,0 +1,69 @@
...
+    value: function(newVal) {
```
rijkvanzanten: You're mixing 

```js
{
  key: function() {}
}
```

and

```
{
  key() {}
}
```

Lets use the second everywhere 🙂

extensions/core/interfaces/map/display.vue
```diff
@@ -0,0 +1,69 @@
...
+      this.viewMap(newVal);
+    }
+  },
+  mounted() {
+    this.viewMap(this.value);
+  },
+  methods: {
+    viewMap(latlng) {
+      this.latlng = JSON.parse(latlng);
```
rijkvanzanten: > JSON.parse(latlng)

The API has a JSON type too. If that's used, you don't have to / can't double parse the value

extensions/core/interfaces/map/display.vue
```diff
@@ -0,0 +1,69 @@
...
+        }
+      }
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+.map-display {
+  width: 200px;
```
rijkvanzanten: Let's use the standardized `var(--width-normal)` here

extensions/core/interfaces/map/display.vue
```diff
@@ -0,0 +1,69 @@
...
+  position: relative;
+  display: inline-flex;
+  flex-direction: column;
+}
+@media only screen and (max-width: 800px) {
+  .interface-map-wrap {
+    display: flex;
+  }
+  .map-display {
+    width: 100% !important;
```
rijkvanzanten: `!important` ⁉️

extensions/core/interfaces/map/input.vue
```diff
@@ -0,0 +1,166 @@
+<template>
```
rijkvanzanten: See the `display.vue` comments

extensions/core/interfaces/map/map.js
```diff
@@ -0,0 +1,109 @@
...
+      marker: null,
+      latlng: {
+        lat: null,
+        lng: null
+      }
+    };
+  },
+  watch: {
+    //TODO: Check if this watcher is required
+    //? Do we need to re-render map when theme changes from interface debugger?
```
rijkvanzanten: While there isn't an interface that sets other interfaces' options yet, there might be in the future.

extensions/core/interfaces/map/map.js
```diff
@@ -0,0 +1,109 @@
...
+      if (this.marker) {
+        if (latlng) {
+          this.marker.setLatLng(latlng).setOpacity(1);
+        } else {
+          this.marker.setOpacity(0);
+        }
+      } else if (this.map) {
+        // Create A Marker Instance
+        let markerIcon = leaflet.icon({
+          //? Should we use base64?
```
rijkvanzanten: What are the alternatives? An `svg` would already be nicer than a `png` 🙂

extensions/core/interfaces/map/meta.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "$t:map",
+  "version": "1.0.0",
+  "datatypes": {
+    "ARRAY": 255,
```
rijkvanzanten: We should use `"JSON": null` instead

extensions/core/interfaces/map/display.vue
theharshin: So now we're not rendering the map on `display`, this is not an issue anymore. Just curious, `this.name` will return component name right? How will it be unique each time if same component is rendered multiple times? 🤔

extensions/core/interfaces/map/display.vue
theharshin: As we do not display map, I don't need to rerender map & update marker. So I'm removing this watcher. Right?

extensions/core/interfaces/map/map.js
theharshin: How did I forget SVG?! 😅I've used inline SVG now.

rijkvanzanten:
This is getting there!

Few small things to do still:
* `readonly` doesn't seem to do anything
* Rename certain options
  * `Zoom` -> `Min Zoom` (so it fits better with `Max Zoom`)
* Remove lat / long options in favor of default_value
  The API already returns a default value for the fields. If there's no default value, the interface shouldn't render the marker
* I'm not sure if we should make the width user adjustable in pixel values. Thoughts on that @benhaynes? Width the upcoming grid system, shouldn't it "just" `width: 100%` and rely on it's parent for the width?
* Grayscale vs color doesn't seem to change anything
* We should use the color `var(--accent)` for the marker instead of `#039be5`

If there's any other things, I'll report back here 😄 

Thanks!

rijkvanzanten:
Thanks @theharshin!

I implemented one small improvement:

We barely use any of leaflets other features (pop-ups, buttons, layer selects etc), but you did include the full CSS file for everything. I copied the parts we actually need which shaved 11KB off of the map interface's size, which makes it load considerably faster 👍

rijkvanzanten:
Thanks @theharshin!

I implemented one small improvement:

We barely use any of leaflets other features (pop-ups, buttons, layer selects etc), but you did include the full CSS file for everything. I copied the parts we actually need which shaved 11KB off of the map interface's size, which makes it load considerably faster 👍

PR #414, opened 2018-09-18
Rating Interface

theharshin:


rijkvanzanten:
Thanks @theharshin! This is awesome 😄 

I've changed the default of the "display" mode to number, as I'm not 100% sure if the huge colored stars will fit nicely in all contexts where `display.vue` is (gonna be) used 🙂

Issue #449, opened 2018-10-01
Clean .gitignore


WellingGuzman:
`.gitignore` needs a clean up. Unused and rules being used, and some new might be missing.

PR #453, opened 2018-10-01
Unifies Git Ignore Files

moebrowne:
This PR unifies all the Git ignore files all into one single file in the root

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,6 @@ config/*
 !config/api_sample.php
 phpunit.xml
 
-public/core/
-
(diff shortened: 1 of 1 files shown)


WellingGuzman:
I didn't hear back from you @moebrowne, so I did this myself to also close #449 (https://github.com/directus/api/commit/e6addb35df15831432586f93ce3a4e8b2b45c10a).

I tried to use this PR as starting point but ended up confusing with what was going on.

Either way thanks for taking your time to make this PR.

PR #458, opened 2018-10-02
Add user full name to forgot password and reset password templates

lex111:


WellingGuzman:
Notice that this had a minor issue, when there's not first or last name it will create an space between `Hey` and the comma (`,`)

PR #476, opened 2018-10-09
Rating Interface: Fix Datatypes

theharshin:
- Changed `datatypes` format in `meta.json`
- Textbox will display default value as '0' instead of blank


rijkvanzanten:
Thanks @theharshin! 😄

PR #533, opened 2018-10-29
File Preview : Image Quality & MIME Type fix.

theharshin:
This will keep the quality of an image as it is & retain the file type.


benhaynes:
Thanks @theharshin – this makes sense. Bummer that the library handles rotation in such a weird way... I can open a ticket on their repo, but it seems like this is the same? https://github.com/fengyuanchen/cropperjs/issues/332

Does the solution at the end of that issue help us at all?

PR #537, opened 2018-10-29
Removing relationship metadata (directus_relations)

daniel-upzzle:
In the dropColumn method of TableService I added a call to a new method on the same service called "removeRelationship" in case the column has any relationship.

The new method is in charge to remove data related to the relationship of the field being removed.

### DIFF
--- a/src/core/Directus/Services/TablesService.php
+++ b/src/core/Directus/Services/TablesService.php
@@ -742,8 +742,12 @@ class TablesService extends AbstractService
         }
 
         if ($columnObject->isManaged()) {
-            $this->removeColumnInfo($collectionName, $fieldName);
-        }
+			$this->removeColumnInfo($collectionName, $fieldName);
+		}
+		
+		if ($columnObject->hasRelationship()) {
+			$this->removeRelationship($columnObject->getRelationship());
+		}
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daniel-upzzle:
It's lack was causing some troubles when adding, removing, readding relations to collections.

benhaynes:
Thanks @daniel-upzzle !!

Could you remind me which issue this resolves? Is it https://github.com/directus/app/issues/1045?

daniel-upzzle:
I think this will resolve the issue you mention. But I've done this because I've found a similar issue.

1) create a many to many relationship or many to one
2) remove the field with that relation (also using the app)
3) create the field again with a new type of relation (or the same in some cases).
4) couldn't reedit the field options nor include items in a relation

I've found that removing manually the implied directus_realtions I could create the fields again without issues.

daniel-upzzle:
I don't know if I have to close this pull request.

benhaynes:
Hey @daniel-upzzle – why was this closed?

daniel-upzzle:
Sorry, I'm new using pull requests, forking etc. I wasn't able to create a new PR aside from this one, because it was in my master branch.

One thing I don't understand is why fetching from the upstream (and rebasing its master to my master) did not eliminate the changes of this PR I have locally when the upstream remote branch being fetch has not this changes applied yet.

rijkvanzanten:
Hey @daniel-upzzle! Is that first commit https://github.com/directus/api/pull/537/commits/b14a933299613c20269e4cd2a2cdf9d931e4d728 the only one you'd like merged? I can clean up the history of this PR a bit but I don't want to delete any other changes you've made 🙂

daniel-upzzle:
Richt @rijkvanzanten , that is the commit making the point.

Thanks!

Issue #542, opened 2018-10-30
Call to undefined method DateTime::toISO8601Format() 


blackdoginet:
# Bug Report

## Steps to Reproduce

1. Apply patches from  commit 7245b7aee01e2362e38b9d79d6cc024c568bf276 to cloned suite due to previos issues
Fixes #531
Ref: #532
Ref: #498

2. Login get error 

## Actual Behavior
![image](https://user-images.githubusercontent.com/17192852/47714795-c4ec5480-dc35-11e8-97a8-3d01503f08f5.png)

## Other Context & Screenshots

public function toISO8601Format() is in DateTimeUtils.php
## Technical Details

- Device: dedicated Kimsufi server
- OS: Debian GNU/Linux 9 (stretch)
- Web Server: Apache/2.4.25 (Debian)
- PHP Version:PHP 7.0.30-0+deb9u1
- Database: 10.1.26-MariaDB
- Install Method: cloned master suite and patched commit 7245b7aee01e2362e38b9d79d6cc024c568bf276

Error log
[...]

PR #547, opened 2018-10-31
Set time format to iso08602 reformulated to prevent server error

daniel-upzzle:
I think it resolves:

https://github.com/directus/api/issues/542

I was having the same or similar problem.

Daniel


WellingGuzman:
Thanks for contributing @daniel-upzzle, but I have fixed this with https://github.com/directus/api/commit/02ac62ffc58290efef1060f20f15603897de329b

PR #552, opened 2018-10-31
Update extensions/core/interfaces/map/input.vue

FrikanRw:
Added getCoordinatesforPlaceName method.
Added input for address to geocode

Not sure about styling will change if helped.
Not sure about setValue usage.

### DIFF
--- a/extensions/core/interfaces/map/input.vue
+++ b/extensions/core/interfaces/map/input.vue
@@ -10,6 +10,12 @@
 
       <div class="map-actions">
         <!-- TODO: Add Geo Search Here -->
+        <input v-model="placeName" placeholder="Enter address to geocode"/>
+        <button
+          v-if="isInteractive"
+          @click="getCoordinatesforPlaceName(placeName)">
+          <i class="material-icons">add_location</i>
+        </button>
         <button
           v-if="isInteractive"
           class="map-my-location"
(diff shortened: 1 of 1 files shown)


extensions/core/interfaces/map/input.vue
```diff
@@ -273,7 +279,23 @@ export default {
...
         this.darkAccentColor
       }" d="M12 2C8.13 2 5 5.13 5 9c0 5.25 7 13 7 13s7-7.75 7-13c0-3.87-3.13-7-7-7zm0 9.5c-1.38 0-2.5-1.12-2.5-2.5s1.12-2.5 2.5-2.5 2.5 1.12 2.5 2.5-1.12 2.5-2.5 2.5z"/><path d="M0 0h24v24H0z" fill="none"/></svg>`;
-    }
+    },
+
+
+
+      // Place name for geocode lookup on openstreetmap database via Nominatim, Returns coordinates in [lat,lon]
+    getCoordinatesforPlaceName(placeName) {
+      axios
```
rijkvanzanten: Isn't `axios` undefined in this case?

rijkvanzanten:
Also @benhaynes we should probably make this opt-in as an interface option. Otherwise we're making outgoing HTTP requests to third parties with sensitive information without the users consent.

benhaynes:
Agree. Can this be added into and Interface Option that defaults to disabled?

rijkvanzanten:
Hey @FrikanRw! The original PR is a good start, but there are a few critical things that need to be fixed / added before this can be merged in. Do you mind taking another look? I can do it as well, but it's gonna take a few weeks as there are more pressing concerns in the main app right now..

* **Styling**  
There seems to be no styling whatsoever, breaking the looks of the map interface:

image

* **Axios is undefined when used**  
Using `this.$axios` instead of `axios` should do it.

* **There's no data key for `placeName`**  
Vue will throw an error when typing in the input, because the property `placeName` doesn't exist in `data()`

* **Adhere to the Nomatim usage policy**  
The Nomatim API has some restrictions on it's usage:

> Requirements
> * No heavy uses (an absolute **maximum of 1 request per second**d).
> * **Provide a valid HTTP Referer or User-Agent** identifying the application (stock User-Agents as set by http libraries will not do).
> * **Clearly display attribution** as suitable for your medium.
> * Data is provided under the ODbL license which requires to share alike (although small extractions are likely to be covered by fair usage / fair dealing).

* **Make it opt-in**  
This feature should be turned off by default and have a warning message noting that we use an external third party service to retrieve the lat long.

FrikanRw:
@rijkvanzanten  I will gladly complete this merge. I have moved in the last two weeks so my life was in boxes. I will implement your changes and try my hand at styling to fit. I will ask questions here if I need anything.

extensions/core/interfaces/map/input.vue
FrikanRw: Yeah I was wondering if I should Fetch or something else?

rijkvanzanten:
Thanks @FrikanRw! Hope your move went well 🙂

rijkvanzanten:
These changes have been updated and implemented by @Kinzi in #637

PR #553, opened 2018-11-01
Add timezone and location for Czech Republic to schema.sql

AntiCZ:
Hello, It is just a suggestion for add timezone and location for the Czech Republic to .sql


rijkvanzanten:
Hey @AntiCZ! Thanks for the PR. You've actually added 'support' for the Czech language in the DB. However, without the corresponding [translations file](https://github.com/directus/app/blob/master/src/lang/locales/en-US.js), the app still can't render text in Czech. I'll have to close this for now until we have full support for the Czech language

PR #564, opened 2018-11-02
use .gitignore instead of .gitkeep for the keep empty folder

alioygur:
.gitkeep keeps logs folder but tracks generated logs files.

### DIFF
--- a/logs/.gitignore
+++ b/logs/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore


WellingGuzman:
Thanks for the contribution, but recently we agreed on using `.gitkeep` instead of `.gitignore` recently to have everything merged in one file in the root directory.

rijkvanzanten:
> to have everything merged in one file in the root directory

What do you mean?

> .gitkeep keeps logs folder but tracks generated logs files.

Git shouldn't be tracking the log files, as every `.log` file is being ignored in `.gitignore` @alioygur

WellingGuzman:
> What do you mean @WellingGuzman?

To have all `.gitignore` file merged in one `.gitignore` in the root directory.

benhaynes:
I think there is some confusion here, all we want to do is keep all of the gitignores in one place... here is where all rules should be added:

https://github.com/directus/api/blob/master/.gitignore

Can we adjust this PR so that it uses the main .gitignore file?

WellingGuzman:
The `.gitignore` in the root ignores every `.log` files which include all the logs written into the `logs` directory.

Ref: https://github.com/directus/api/blob/master/.gitignore#L3

benhaynes:
Ahh, so this should be working already. And since I am not being asked to commit any log files locally I think it is. 

@alioygur – are you seeing any issues here that we are unaware of? I'll close this in the meantime

alioygur:
> @alioygur – are you seeing any issues here that we are unaware of? I'll close this in the meantime

Recently i have cloned directus/api build branch, this branch doesn't have the .gitignore file in the root directory so the logs files tracked by git.

benhaynes:
Ahh, that makes more sense! Just needed a little context. We accidentally were stripping the root `.gitignore` files during the build process... but we will NOT do that any more. `7.0.5/2.0.5` will be released in a few minutes and will still have the `gitignore` in it... so you should be all set!

Thanks for clarifying!

Issue #560, opened 2018-10-30
Background jobs (Push Notifications)


vzool:
Hi,

I'm looking for a way to do some long jobs in the background like (Custom mail sending,  SMS calls, Push Notifications, Reports generation and any Remote Service Calls). It there a way to make this and I'm not aware of?

Thanks

rijkvanzanten:
Ideally, we'd have a system where you can have custom snippets of PHP (similar to hooks) that can be run at an interval. If these snippets have access to the Directus internal api's (like hooks), you should be able to do virtually whatever you want.

I think we should / could have a `directus_tasks` collection that keeps track of at what interval to run what function? And maybe use `directus_activity` for logs of said task?

benhaynes:
I like that. Would the code to be executed be saved in the database (eg `directus_tasks.code`)? Would it be a reference to a file (maybe uploaded in Directus Files) so that it is language agnostic?

If the tasks define when it runs (start, frequency, end/iterations), and we have a simple CRON that fires every minute, then we can just check the tasks and run them as needed. Seems pretty straightforward.

rijkvanzanten:
That could be even easier @benhaynes.

If the file itself starts with something to tell the API how often to run, we don't need to the database in between. We could have an `extensions/tasks` folder with `.php` files that get fired

benhaynes:
Interesting. So naming could use [cron schedule expressions](https://crontab.guru/#5_4_*_*_sun):

### “At 04:05 on Sunday.”
```
/extensions/tasks/[5_4_*_*_sun]task-name.php
```

We could also "disable" tasks just like other extensions, by pre-pending `_` to the file-name.

Now this is starting to look like a nice clean solution that just needs a CRON parser to run/check every minute!

benhaynes:
https://github.com/directus/api/pull/567

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

PR #567, opened 2018-11-03
Add cron worker for dispatched jobs through the database

vzool:
Hi, 

This is a cron job worker using the database as a transfer layer between the dispatcher and the consumer which is a cron job in our situation. and it does resolve the feature request #560 

- First: initialize the worker by:

```bash
> bin/cron_worker init # this will create a new table called `enqueue`
```

- Second: call worker directly and it will consume any dispatched jobs

```bash
> /home/app/directus/bin/cron_worker
```

Any Task **should** implements [TaskInterface](https://github.com/vzool/api/blob/5461735d66c5d3b350aa42c0599535c514c5cd29/src/core/Directus/Task/TaskInterface.php) like this:

```php

namespace Directus\Customs\Tasks;

use Interop\Queue\PsrMessage;
use Interop\Queue\PsrConsumer;
use Directus\Task\TaskInterface;

class HelloWorld implements TaskInterface{

 	public function handle(PsrMessage $message, PsrConsumer $consumer){

                // $message->getProperties() 
		// $message->getHeaders()
		
		// process a message
		
		$consumer->acknowledge($message);
		// $consumer->reject($message);
	}	
}
```

**Testing Progress**:

- [x] Local: direct interaction test at vagrant box(Ubuntu 18.04.1 LTS). [**Done**]
- [x] Local: cron test at vagrant box(Ubuntu 18.04.1 LTS). [**Done**]
- [x] Production: direct interaction test at Debian 9. [**Done**]
- [x] Production: cron test at Debian 9. [**Done**]

**Notes**:

- If there is a failed or bad job in the queue, it will stop that queue and the worker will ignore the queue until you resolve the issue.
- Worker support multiple queues, so if job failed in one that will not affect the other.
- If a job failed the worker will requeue it again, exits the current queue and move to other queue looking for jobs in it.
- The worker runs in `Exit-On-First-Error` mode by default which make sense for cron jobs environment if you called every minute, but you can change that behaviour by the following line:

```bash
> /home/app/directus/bin/cron_worker ignore-errors 
```

- If you enabled `Ignore-Errors` mode and there is an error the worker will hit that error every single second until you fix it. This is so annoying if you have an automatic bug reporter by email or SMS.

Any idea is welcome here, Thanks 😘

### DIFF
--- a/bin/cron_worker
+++ b/bin/cron_worker
@@ -73,7 +73,7 @@ set_error_handler(function(int $errno , string $errstr){
 $task_class_namespace = $config['tasks']['namespace'];
 
 // Check direcory selected for the Tasks
-$class_dir_path = $config['tasks']['path'];
+$class_dir_path = BASE_PATH . DS . $config['tasks']['path'];
 
 if(!file_exists($class_dir_path)){
 	
@@ -108,7 +108,7 @@ foreach($config['tasks']['queues'] as $queue => $tasks){
 
 					try{
 
-						$class_full_path = BASE_PATH . DS . $class_dir_path . DS . $class_path . '.php';
+						$class_full_path = $class_dir_path . DS . $class_path . '.php';
 
 						if(!in_array($class_path, $loaded_class)){
 


vzool:
😊

vzool:
@WellingGuzman Finally, I guess I did all the tests now 😆

vzool:
@WellingGuzman Still reviewing this PR 🙃

benhaynes:
@WellingGuzman — can you post some thoughts on this here?

philleepflorence:
Is it possible for this to be bundled as an extension, also can the installer specify the name of the collection that would hold the queue?

benhaynes:
Hey @vzool — Welling left the organization, but we'll have @bjgajjar take a look.

@philleepflorence — I would say that this should be in in the Core (not an extension)... as long as we find a modular way to allow for different server schedulers (Apache, nginx, etc). I think we could have a single system collection for managing the job queue since that is not server dependent. Something like `directus_tasks`:

* `id`
* `title`
* `frequency` (either in raw minutes or with a cron-like syntax: https://en.wikipedia.org/wiki/Cron)
* `script` (maybe a relational link to a file? If Directus Files we'd need a way to make these private)

Then we just need a modular way to run the code every minute. My thinking is that we have a cron job (or whatever your server supports) that runs every minute, checks the `directus_tasks` for anything that needs to run, and runs the script file for whatever has a matching frequency. 

I'm not sure how this PR works... but that's my (super basic) concept. :)

philleepflorence:
@benhaynes - agreed, makes sense.
The script could be a custom or core endpoint.
My use case on my custom cronjob build is syncing data between social media and directus DB via an endpoint.

benhaynes:
Interesting, I like the idea of allowing the task to hit an endpoint. My only concern is that there's no way to setup custom endpoints through the App, which means you need server access. The benefit of the "file" approach is that we could allow for uploading and setting up everything through the app. So perhaps we look into supporting both?

benhaynes:
@rijkvanzanten and @bjgajjar — could you guys give me your thoughts on this PR? I like the idea of having a way to schedule tasks, but I don't know enough to audit this code/approach.

rijkvanzanten:
I like the idea of being able to do stuff at a given time, but it's unfortunate that you have to go "around" PHP (eg do stuff on the CLI) to set it up. It makes sense, as PHP doesn't have a constant running process, but that does make it a bit hard to use.

philleepflorence:
One way to do it is to have a master Cron Process per installation that runs every minute and then manage project tasks via directus, having PHP call the scripts or URLs of the individual tasks.

On cronjobs collection save in the DB, update a file in the filesystem that the cronjob looks at - minimizes DB overhead.

This is a set up currently in use on some Directus installations. So there is no need for system admins after launch.

benhaynes:
I like the idea of having this be "decoupled from PHP" as much as possible. I'd love to store the tasks in the database so that if we move to node.js API in the future the tasks themselves don't need to change.

And having a CRON job that runs every minute (like @philleepflorence mentioned) and then check/trigger the tasks as needed sounds perfect. I like it!

Issue #571, opened 2018-11-04
Error when building the `npm` dependencies of the extensions on Windows OS


sayak-brm:
# Bug Report

## Steps to Reproduce

Build the `npm` dependencies of  the extensions on Windows OS using `npm run build` after `npm install`.

## Expected Behavior

The extensions are built without any error.

## Actual Behavior

The build stops due to the use of *nix specific shell commands.

## Other Context & Screenshots

![image](https://user-images.githubusercontent.com/5992716/47969670-19ae2600-e0a1-11e8-8a18-8a080095a57f.png)

## Technical Details

- API: Directus API 2.0.5
- OS: Windows 10 Home x64 ver 1803
- Web Server: Apache 2.4.33
- PHP Version: PHP 7.2.6
- Node.js Version: 10.13.0 (x64)
- Install Method: cloned `master` branch

## Log File

```
0 info it worked if it ends with ok
1 verbose cli [ 'D:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'D:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'run',
1 verbose cli   'clear' ]
2 info using npm@6.4.1
3 info using node@v10.13.0
4 verbose run-script [ 'preclear', 'clear', 'postclear' ]
5 info lifecycle @directus/extensions-builder@1.0.0~preclear: @directus/extensions-builder@1.0.0
6 info lifecycle @directus/extensions-builder@1.0.0~clear: @directus/extensions-builder@1.0.0
7 verbose lifecycle @directus/extensions-builder@1.0.0~clear: unsafe-perm in lifecycle true
[...]

PR #572, opened 2018-11-04
Fix issue #571 in directus/api

sayak-brm:
Replaced *nix shell commands with a cross-platform Node.js alternative [rimraf][1] so that the extensions can be built on non-*nix platforms (eg. Windows).

  [1]: https://www.npmjs.com/package/rimraf


rijkvanzanten:
Thanks @sayak-brm! 😄

benhaynes:
Thank you so much @sayak-brm !! We really appreciate the contribution!

PR #589, opened 2018-11-15
Set port based off of scheme if host doesn't include port

Anthonyzou:
When directus api will assume the port is 80 even when the host string doesn't include a port but the scheme is https.

This adds a quick check for that case.

### DIFF
--- a/src/helpers/all.php
+++ b/src/helpers/all.php
@@ -169,12 +169,16 @@ if (!function_exists('create_uri_from_global')) {
 
             if (isset($matches[2])) {
                 $port = (int)substr($matches[2], 1);
+            } else {
+                $port = $isSecure ? 443 : 80;
             }
         } else {
             $pos = strpos($host, ':');
             if ($pos !== false) {
                 $port = (int)substr($host, $pos + 1);
                 $host = strstr($host, ':', true);
+            } else {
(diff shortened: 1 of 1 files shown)


src/helpers/all.php
```diff
@@ -169,12 +169,16 @@ function create_uri_from_global()
...
             }
         } else {
             $pos = strpos($host, ':');
             if ($pos !== false) {
                 $port = (int)substr($host, $pos + 1);
                 $host = strstr($host, ':', true);
+            } else {
+                $port = $isSecure ? 443 : 80;
             }
         }
```
WellingGuzman: I have no way to test this out at the moment, but I believe in this case it will be best to check the condition outside.

Example: 
```
if ($isSecure && $port === 80) {
    $port = 443;
}
```

AntiCZ:
if ($isSecure && $port === 80) {
    $port = 443;
}

I do know no if this is the solution. In the last two years, we are "fighting" with sysadmins about what application should do and want should do not. The conclusion is two things. Check if is secure if app generate absolute links (http://domain.com or https://domain.com) and works in sub dir and everything else is problem sysadmins.

With this is "if" is one problem. What if sysadmin listens https on 80, It is stupid but the application should do not care about it. This is a sysadmin problem.

PR #600, opened 2018-11-20
UX fix - Better experience for user if upload .pdf

AntiCZ:
Hello, PR solves a problem when the user uploads successfully .pdf.

Step to reproduce. Upload .pdf file. Everything is successful but user see broken icon and can not clicks on any link to check if .pdf is OK.

Also works for a video icon. I do not know if you would like to link to video too.

### DIFF
--- a/extensions/core/interfaces/file/get-icon.js
+++ b/extensions/core/interfaces/file/get-icon.js
@@ -1,4 +1,9 @@
 export default function getIcon(type) {
+
+  if (type === "application/pdf") {
+    return "picture_as_pdf";
+  }
+
   if (type.startsWith("application")) {
     return "insert_drive_file";
   }
--- a/extensions/core/interfaces/file/input.vue
+++ b/extensions/core/interfaces/file/input.vue
@@ -5,7 +5,9 @@
       class="card"
       :title="value.title"
       :subtitle="subtitle"
-      :src="value.data.full_url"
+      :src="src"
+      :icon="icon"
+      :href="href"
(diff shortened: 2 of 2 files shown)


benhaynes:
Thanks @AntiCZ !! @rijkvanzanten will take a look at this one since it's an interface (technically the App). 

We really appreciate the PR! This could potentially solve this ticket too:

https://github.com/directus/app/issues/729

AntiCZ:
If it's ok. I'll also fix the `extensions/core/layouts/cards/layout.vue`

benhaynes:
That'd be great @AntiCZ !

I haven't reviewed this yet, but it seems like the ideal solution would be to show the file extension if a thumbnail isn't available. Is that what this PR is doing?

AntiCZ:
@benhaynes Yes, you are right.

Right now. Problem is that file src try to be an image. `` This is fails and broken icon shows up. PR sends src to an  only if mime-type starts with "image". If is apilication or video so src is null so no error rise up and be shown icon at https://github.com/directus/app/blob/master/src/components/card.vue line:36

But PR handles form input file on the item detail. If @rijkvanzanten says it is good I would like to also fix previews in File library but this is more complicated.

rijkvanzanten:
Looks good to me! Thanks @AntiCZ 😄

rijkvanzanten:
The previews in the file library might look more difficult on the outside, but they use the same card components so I feel like the fix is very similar :)

Issue #594, opened 2018-11-17
Remove scope from directus_settings


rijkvanzanten:
# Bug Report

We don't use the `scope` column in directus_settings anymore. I'm not a 100% sure how the migrations have to be setup to get this done correctly.

PR #605, opened 2018-11-22
Fixed database installer

marcomontalbano:
Hi, when you try to change the `color` from the APP in `Global Settings` you get an error:

schermata 2018-11-23 alle 00 29 00

I only added the default property to the installer.


rijkvanzanten:
Hey @marcomontalbano! This doesn't fix the scope issue (#594), but it's nice to have the defaults in the DB either way :) 

The only thing that's different is that the color we use is actually `light-blue-600`, not `400` 🤔

PR #607, opened 2018-11-23
Remove scope from directus_settings

### LINKED ISSUE #594: Remove scope from directus_settings

rijkvanzanten:
Adds a migration to remove the scope column.

Closes #594, ref https://github.com/directus/app/issues/1140

### DIFF
--- a/migrations/db/schemas/20180220023243_create_settings_table.php
+++ b/migrations/db/schemas/20180220023243_create_settings_table.php
@@ -29,10 +29,6 @@ class CreateSettingsTable extends AbstractMigration
     {
         $table = $this->table('directus_settings', ['signed' => false]);
 
-        $table->addColumn('scope', 'string', [
-            'limit' => 64,
-            'default' => null
-        ]);
         $table->addColumn('key', 'string', [
             'limit' => 64,
             'null' => false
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
@WellingGuzman mentioned on Slack that there are some places in the API code that relies on `scope` being there. Merging this now would cause issues as the code will be trying to access a column that doesn't exist

rijkvanzanten:
@WellingGuzman Did you remove all other references to `scope` too?

andreekeberg:
All references are definitely not removed as I just did a `git pull` and still an error that "**scope: This value should not be blank.**" when trying to set a "Color" in the Global Settings.

rijkvanzanten:
@andreekeberg Have you ran the DB update with the `/update` endpoint or by using the button in settings?

andreekeberg:
@rijkvanzanten Yes, I just found the button and tried this, and still got the same error.
I'm wondering if it could in some way be related to the bug that makes it impossible to select a logo, because of some kind of database migration. Both issues persist after the DB update.

WellingGuzman:
Hey @andreekeberg, it seems to be an API error: `scope: This value should not be blank.`. Is the `directus_settings` table still have the `scope` column?

andreekeberg:
Yes @WellingGuzman the `directus_settings` table still has this column after the DB update.

WellingGuzman:
@andreekeberg Are you using the `master` branch of `directus/api` repository? This fix hasn't been released yet.

PR #618, opened 2018-11-29
S3 adapter request header options

roncodes:
This is a small PR to add in S3 request header options such as ACL via the storage config


WellingGuzman:
@roncodes I renamed the options from `s3_options` to `options`, to make this possible in a future to be shared with any other adapters.

Issue #576, opened 2018-11-07
Incorrect query for "all" and "has" filter


wellingguzman:
The `all` and `has` filter is generating a bad query.

beac0n:
Also, using something like filter[locatios.id][eq]=1, where locations is a many-to-many relation, also leads to a bady query.

@WellingGuzman : I'd like to help you with this, as I have run into this issue myself multiple times.
 I could provide a pull request, if you would point me into the right direction, where to look in the codebase :-)

WellingGuzman:
@beac0n what's the bad query you are getting?

To point you to the right direction, the filters are created here in the [`doFilter`](https://github.com/directus/api/blob/master/src/core/Directus/Database/TableGateway/RelationalTableGateway.php#L1281) method.

I don't know where is the actual problem, but here is the [QueryBuilder](https://github.com/directus/api/blob/master/src/core/Directus/Database/Query/Builder.php) this is the object that creates the query based on the filters passed from the `doFilter` method.

sebj54:
Hi!
I'm having this problem too! And I'd like to help too :)

First of all, what's the correct query? As I never made it work, I'm not sure what I'm doing is correct.

I have a master collection `inspirations` which has a many-to-many field called `saisons` (sorry for French names, I had to!). It is linked with the `saisons` table via the ìnspirations_saisons` associative table.

So we have:
```
----------------   ------------------------   -----------
| inspirations |   | inspirations_saisons |   | saisons |
|--------------|   |----------------------|   |---------|
|              |   | id                   |   |         |
| id           | ← | inspiration          |   |         |
|              |   | saison               | → | id      |
----------------   ------------------------   -----------
```

If I want to search for all inspirations with 1 as ID, what should be my API query?

## First attempt: `saisons.saison.id`
`http://local.directus.com/_/items/inspirations?fields=*.*,saisons.saison.*&filter[saisons.saison.id][has]=1`

When I hit this URL, I get the following error:
```
{
    "error": {
        "code": 202,
        "message": "Unable to find field \"saison\"",
        "class": "Directus\\Database\\Exception\\FieldNotFoundException",
        "file": "C:\\wamp64\\www\\directus\\src\\core\\Directus\\Database\\SchemaService.php",
        "line": 333
    }
}
```

## Second attempt: `saisons`
`http://local.directus.com/_/items/inspirations?fields=*.*,saisons.saison.*&filter[saisons][has]=1`

When I hit this URL, I get the following error:
```
{
    "error": {
[...]

PR #619, opened 2018-11-29
Add support for filters on a many-to-many relation

sebj54:
Hi!

This is my first contribution so I'm not sure how my changes affect all functionalities. So I'm really open to your thoughts on my changes, I'm really happy to be able to contribute to this awesome project! 😃 

So I added support for filters on many-to-many relations. This is maybe a fix for this issue: https://github.com/directus/api/issues/576. Even if it does not use `has` but `in`, it is the solution to my initial need (I'm not sure now if `has` was the solution).

I used the same tables as I described in the issue:

```
--------   --------------   ---------
| test |   | test_types |   | types |
|------|   |------------|   |-------|
|      |   | id         |   |       |
| id   | ← | test       |   |       |
|      |   | type       | → | id    |
--------   --------------   ---------
```

I tested my changes with this query: `http://local.directus.com/_/items/test?fields=*.*&filter[types.id][in]=1`. I also tested with another table and two similar filters on associative tables and it works too!

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -20,6 +20,7 @@ use Directus\Util\ArrayUtils;
 use Directus\Util\DateTimeUtils;
 use Directus\Util\StringUtils;
 use Zend\Db\Sql\Expression;
+use Zend\Db\Sql\Predicate\In;
 use Zend\Db\Sql\Predicate\PredicateInterface;
 use Zend\Db\Sql\Select;
 use Zend\Db\Sql\Sql;
@@ -948,7 +949,13 @@ class RelationalTableGateway extends BaseTableGateway
         }
 
         if (in_array('total_count', $list)) {
-            $metadata['total_count'] = $this->countTotal();
+            $condition = null;
+            if ($this->getTableSchema()->hasStatusField()) {
+                $fieldName = $this->getTableSchema()->getStatusField()->getName();
+                $condition = new In($fieldName, $this->getNonSoftDeleteStatuses());
+            }
+
+            $metadata['total_count'] = $this->countTotal($condition);
         }
 
         if ($tableSchema->hasStatusField() && in_array('status', $list)) {
@@ -2097,27 +2104,16 @@ class RelationalTableGateway extends BaseTableGateway
         $statement = $sql->prepareStatementForSqlObject($select);
         $results = $statement->execute();
 
-        $statusMap = $this->getStatusMapping();
         $stats = [];
         foreach ($results as $row) {
-            if (isset($row[$statusFieldName])) {
-                foreach ($statusMap as $status) {
-                    if ($status->getValue() == $row[$statusFieldName]) {
-                        $stats[$status->getValue()] = (int) $row['quantity'];
-                    }
-                }
-            }
+            $stats[$row[$statusFieldName]] = (int) $row['quantity'];
         }
 
-        $vals = [];
+        $statusMap = $this->getStatusMapping();
         foreach ($statusMap as $value) {
-            array_push($vals, $value->getValue());
-        }
-
-        $possibleValues = array_values($vals);
-        $makeMeZero = array_diff($possibleValues, array_keys($stats));
-        foreach ($makeMeZero as $unsetActiveColumn) {
-            $stats[$unsetActiveColumn] = 0;
+            if (!isset($stats[$value->getValue()])) {
+                $stats[$value->getValue()] = 0;
+            }
         }
 
         return $stats;


rijkvanzanten:
This is awesome @sebj54. Thank you so much!

benhaynes:
Wow!! Thank you @sebj54 — what an awesome contribution!! We'll have @WellingGuzman take a look today.

src/core/Directus/Database/Schema/Object/FieldRelationship.php
```diff
@@ -109,6 +120,7 @@ public function isOneToMany()
     protected function guessType()
     {
         $fieldName = $this->fromField->getName();
+        $interface = $this->fromField->attributes->get('interface');
```
WellingGuzman: This is a good idea, but this will only be attached to an interface, and not the `m2m` type (which doesn't exists). In this version we are trying to avoid attaching logic based on interface names in the core unless it's an extension.

sebj54:
Ok I see what you're pointing out. I was wondering if I was not doing something not intended when I saw there was no getter for interface.

Actually I think my changes are still needed because this filter does not work out of the box: `filter[types.id][in]=1`. 

So if I follow the new logic, I should change conditions in my code to test on O2M relation rather than M2M interface. But is it possible to achieve the same result?

Actually the generated query is this one:

```sql
SELECT `test`.`id` AS `id`, (...)
FROM `test` 
WHERE `test`.`types` IN (
    SELECT `test`.`id` AS `id` FROM `test` WHERE `test`.`id` IN ('1')
)
ORDER BY `test`.`id` IS NULL, `test`.`id` ASC, `test`.`id` ASC LIMIT 200 OFFSET 0
```

and it should be (approximately) this one:

```sql
SELECT `test`.`id` AS `id`, (...)
FROM `test` 
WHERE `test`.`id` IN (
    SELECT `test_types`.`test` AS `id` FROM `test_types` WHERE `test_types`.`type` IN ('1')
)
ORDER BY `test`.`id` IS NULL, `test`.`id` ASC, `test`.`id` ASC LIMIT 200 OFFSET 0
```

Without this part of my changes: https://github.com/directus/api/pull/619/files#diff-ae28a4c569c7bdcc3c0bd6fef223a1a7R1241, I can't access the foreign key field on the associative table which. As this field is specific to many-to-many relations, I'm not sure if I can locate without looking at the many to many relations. 
Have you an hint on how to achieve this same behavior without the check on the many-to-many relation?
Maybe I can check if the field on the main table is O2M, look at the linked table, see if there is a M2O field on it which is pointing to the main table?

src/core/Directus/Database/Schema/Object/FieldRelationship.php
```diff
@@ -101,6 +102,16 @@ public function isOneToMany()
...
     }
 
+    /**
+     * Checks whether the relatiopship is ONE TO MANY
+     *
+     * @return bool
+     */
+    public function isManyToMany()
+    {
+        return $this->getType() === static::MANY_TO_MANY;
```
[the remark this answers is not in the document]
rijkvanzanten: What @WellingGuzman is trying to say is that we treat a many to many as a many-to-one _and_ a one-to-many. There's is no single M2M, but instead there's two (M2O-O2M)

src/core/Directus/Database/Schema/Object/FieldRelationship.php
```diff
@@ -127,7 +139,7 @@ protected function guessType()
             $this->getFieldOne()        === $fieldName &&
             $this->getCollectionOne()   !== null
         ) {
-            $type = static::ONE_TO_MANY;
+            $type = ($interface === 'many-to-many') ? static::MANY_TO_MANY : static::ONE_TO_MANY;
```
rijkvanzanten: To reiterate Welling's previous statement, while this would work just fine, it would lock the functionality down to the interface called `many-to-many`. There will most likely be more many to many style interfaces with different input output formats (think tags, or a specialized "favorites" interface)

WellingGuzman:
> Actually I think my changes are still needed because this filter does not work out of the box: `filter[types.id][in]=1`.

Absolutely, this must be done. I am not saying what you are trying to accomplish is not acceptable, but what I am saying is that how you are trying to do it, it's not how we want it, because we wanted to remove M2M.

> Without this part of my changes: https://github.com/directus/api/pull/619/files#diff-ae28a4c569c7bdcc3c0bd6fef223a1a7R1241, I can't access the foreign key field on the associative table which. As this field is specific to many-to-many relations, I'm not sure if I can locate without looking at the many to many relations.

It used to be similar before, but when the M2M was removed those lines were too. This is a tricky problem now because how can we tell which one is the related O2M? a collection (the junction in this case) could have multiple O2M.

In reality there's not actual M2M, what you could do is this: `types.type.id`, that one should work as you are specifying the junction field (o2m).

sebj54:
> It used to be similar before, but when the M2M was removed those lines were too. This is a tricky problem now because how can we tell which one is the related O2M? a collection (the junction in this case) could have multiple O2M.

Okay, I see, there could be many fields with a O2M relation. Setting the O2M field should work like you said. I'll just have to find a way to get the junction field for the M2O relation with the main table. 

I'll try to change this this week or next week, I'll let you know!

WellingGuzman:
Thanks @sebj54, sorry for the delay, after a good mind refresh this past days, I realized we have a `junction_field` that tells you in that relationship what's the field in the "junction" that goes to the many to one.

Probably you were aware of this, but I am starting to think this can be doable, with less pain than I was thinking before.

sebj54:
I did not mean to close this issue, I removed changes in my master branch to add them on a "feature" branch. Is there a way to reopen this issue by binding the new branch? :(

benhaynes:
Hmm, I don't see a way from my side... but maybe you can re-submit it as a new one?

screen shot 2019-01-21 at 1 37 17 pm

sebj54:
I think I have to push on master again to keep this PR. I'll do this as soon as possible!

amacgregor:
Was this ever merged or implemented? I been trying to get filtering by M2M to work and so far is not possible.

benhaynes:
Hey @sebj54 — any thoughts on this? We'd love to re-open and keep it moving now that you understand what we're after!

benhaynes:
I'm tracking this issue here now since it doesn't seem to be resolved:

https://github.com/directus/api/issues/784

Issue #610, opened 2018-11-24
Support Bundled Extensions (Packages)


benhaynes:
Right now we include extensions within individual directories (interfaces, pages, layouts, etc). It would be nice to have the option to include "packages", or directories that include different extensions that all  serve the same purpose.

For example, if we had a "Project Management" package it might include a kanban board layout, some progress interfaces, and a specific storage adapter. 

It would be nice if all of these extensions could stay grouped within a parent directory instead of "unpacked" into their respective extension directories.

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

PR #623, opened 2018-12-02
Support Bundled Extensions (Packages)

mikefaust:
Related to #610 

Added support for bundled extension directories. They can be added next to the default `custom` bundle directory, e.g.:

- `extensions/custom`
- `extensions/my-first-extension`
- `extensions/project-management`

If a directory name contains multiple words, then separate them with a hyphen. 

As usual the bundled extension directory contains the directories for `endpoints`, `hooks`, `interfaces`, `layouts`, `pages` and co.

Bundled extensions that should be activated must be added to the `config/api.php` file:

````php
// These extensions will be loaded
'extensions' => [
    'custom',
    'my-first-extension',
    'project-management',
],
````

For autoloading to work, the bundled extensions also must be added to `composer.json`:

````json
"autoload": {
    "psr-4": {
        "Directus\\Extensions\\Custom\\": "public/extensions/custom/",
        "Directus\\Extensions\\MyFirstExtension\\": "public/extensions/my-first-extensions/",
        "Directus\\Extensions\\ProjectManagement\\": "public/extensions/project-management/"
    }
}
````

Everything else works as usual. 

The only secret I could not solve how to handle migrations of bundled extensions. If I call the `/_/update` route the default migrations from `upgrades/schemas/*.php` are executed and the paths from `config/migrations.php` are overridden in the `InstallerUtils::getMigrationConfig()` method. Maybe someone can point me in the right direction?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,7 @@ composer.lock
 !/public/extensions/custom/endpoints/_example.php
 
 ##  Hashers
-!/public/extensions/custom/hashers/_CustomHasher.php
+!/public/extensions/custom/hasher/_CustomHasher.php
 
 ##  Hooks
 !/public/extensions/custom/hooks/_products
--- a/.gitignore
+++ b/.gitignore
@@ -24,16 +24,11 @@ composer.lock
 /phpunit.xml
 /documents
 
-# Core extensions
-/public/extensions/core/interfaces
-/public/extensions/core/layouts
-/public/extensions/core/pages
-
-# Custom extensions
-/public/extensions/custom/*/*
+# Extensions
+/public/extensions/*/*
 
 ##  Auth
-/public/extensions/custom/auth
+!/public/extensions/core/auth
 
 ##  Endpoints
 !/public/extensions/custom/endpoints/_directory/
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,6 @@
       "Directus\\": "src/core/Directus/",
       "Directus\\Authentication\\Sso\\Provider\\": "public/extensions/core/auth",
       "Directus\\Api\\Routes\\": "src/endpoints/",
-      "Directus\\Custom\\Embed\\Provider\\": "public/extensions/custom/embeds/",
       "Directus\\Extensions\\Custom\\": "public/extensions/custom/"
     },
     "files": [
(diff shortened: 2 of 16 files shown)


rijkvanzanten:
Woa this is awesome @mikefaust!! Is there anything we can do to negate the composer.json step? Without that, it's a truly drag and drop install which is super easy. Having to change a file is an extra hurdle

mikefaust:
> Woa this is awesome @mikefaust!! Is there anything we can do to negate the composer.json step? Without that, it's a truly drag and drop install which is super easy. Having to change a file is an extra hurdle

Yes we could skip this step if we just add the `extensions` directory to `composer.json` and rename the bundled extension directories to UpperCamelCase to get the full autoloading experience.

````json
"autoload": {
    "psr-4": {
        "Directus\\Extensions\\": "public/extensions/"
    }
}
````

So instead of `public/extensions/my-first-extension` as directory name we would need to use `public/extensions/MyFirstExtension`. Which would also mean renaming the `public/extensions/core` and `public/extensions/custom` directories to UpperCamelCase for clarity and standardization.

rijkvanzanten:
@WellingGuzman @benhaynes is that something we can change for the core ones? The ability to FTP up a single folder to have a collection of different extensions is amazing and IMO crucial for less tech-savvy users.

benhaynes:
Yeah, so the change is all extension directories need to be camel case? Even core ones?

mikefaust:
> Yeah, so the change is all extension directories need to be camel case? Even core ones?

They don't need to be. But imagine new developers trying to create their own extensions. They will look at the core and custom extensions and copy the file structure. So for consistency I would recommend to rename it.

benhaynes:
@mikefaust — any thoughts on Welling's last comment? Our Stale bot is trying to kill this PR but I'd love to keep it alive if we can finish it up!

benhaynes:
Hey @rijkvanzanten — any thoughts on this? We should discuss if we want this... or if it's too outdated at this point to merge in.

rijkvanzanten:
I really like the idea of being able to package up multiple extensions, could be really useful for the future marketplace too! (F.e. having a single bundle that has a custom page, 3 interfaces, and 2 hooks )

We do have to think through how it fits in with support for hooks and endpoints within other extensions though; right now pages can register their endpoints within the same extension folder too

PR #628, opened 2018-12-04
Check version variable for MariaDB as well as version_comment

assmith:
On AWS at least, _version_comment_ contains the rather unhelpful value "Source distribution". _version_ however contains "10.3.8-MariaDB-log".  This PR is for a suggested widening of the variable search to include _version_ when testing for MariaDB.


benhaynes:
Thank you @assmith! All these PRs are so helpful! We'll have @WellingGuzman review them tomorrow morning.

Really appreciate the contributions. 😄

PR #629, opened 2018-12-04
Remove compound key (‘key’,’scope’) before removing scope column

assmith:
I found that the compound index previously on ('key','scope') prevented the scope column from being removed during this migration.  This PR suggests removing the key then deleting the column then recreating the key as per the current schema in 20180220023243_create_settings_table.

As per the current schema I've left the index name as _idx_scope_name_, although maybe it should just be called _idx_name_ now?

### DIFF
--- a/migrations/upgrades/schemas/20181123171520_remove_scope.php
+++ b/migrations/upgrades/schemas/20181123171520_remove_scope.php
@@ -13,7 +13,7 @@ class RemoveScope extends AbstractMigration
               ->save();
         $table->addIndex(['key'], [
             'unique' => true,
-            'name' => 'idx_scope_name'
+            'name' => 'idx_key'
         ]);
     }
 }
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Thanks @assmith!

I think you're right. We should probably call the new index just idx_key instead of having the word scope in there (to prevent future confusion). Also, let's call it idx_key instead of idx_name 🙂

assmith:
Sounds good, I wasn't sure if there was an internal reason for using _name_, _key_ would be better though.

migrations/upgrades/schemas/20181123171520_remove_scope.php
```diff
@@ -8,7 +8,12 @@ class RemoveScope extends AbstractMigration
     public function up()
     {
         $table = $this->table('directus_settings');
+        $table->removeIndexByName('idx_scope_name');
         $table->removeColumn('scope')
               ->save();
+        $table->addIndex(['key'], [
+            'unique' => true,
+            'name' => 'idx_scope_name'
```
WellingGuzman: We can change this to `idx_key

WellingGuzman:
Thanks @assmith!

PR #637, opened 2018-12-07
Update map interface

Kinzi:
Added an address input for finding geocodes to the map and implemented change requests based on https://github.com/directus/api/pull/552

Added a toggle in the options to make this optional. Also added link to Nominatim for credits.

I used bottleneck as a limiter to keep requests under 1 sec - not sure if this is really needed though as it's hard to make more than 1 request / second manually.

Not sure how to solve the referrer/user-agent thing so I didn't tackle that.


benhaynes:
Thank you @Kinzi !! We'll review this and get it merged ASAP!

rijkvanzanten:
The browser will use the URL as referer and the browser as user-agent string by default, so we're all good 👍

PR #644, opened 2018-12-10
Edit file modal for files interface

Kinzi:
Hi guys,

updated the new files interface to feature an edit modal like the M2M interface. You can open it from the card menu of the file

Seems to work well - just a little CSS bug with the tooltip popover `z-index`being above the modal. But I think this would better be fixed by closing the popover `@click` altogether?!?

Let me know if you need anything else! :)

### DIFF
--- a/extensions/core/interfaces/files/input.vue
+++ b/extensions/core/interfaces/files/input.vue
@@ -2,7 +2,7 @@
   <div class="input-single-file">
     <div v-if="value" class="preview">
       <v-card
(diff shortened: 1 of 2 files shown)


benhaynes:
Amazing, thank you so much for this contribution @Kinzi!

rijkvanzanten:
I added a handler for the popover buttons that will close the popover in the card. Thanks for the suggestion @Kinzi and thanks for yet another great PR 😄

PR #647, opened 2018-12-11
datetime option to set current datetime

c-ciobanu:
This pr adds the possibility to set the default value of the datetime as the current datetime.
This is relative to the issue https://github.com/directus/app/issues/1127 in the app repo.


benhaynes:
Thank you @c-ciobanu!!! What a great option to have. 😄 

@rijkvanzanten is pretty busy this week but we'll review this for the next release!

PR #649, opened 2018-12-11
WYSIWYG file upload

Kinzi:
Optionally upload files from WYSIWYG to directus instead of storing them Base64 encoded.

Closes https://github.com/directus/app/issues/758

### DIFF
--- a/extensions/core/interfaces/files/display.vue
+++ b/extensions/core/interfaces/files/display.vue
@@ -1,6 +1,8 @@
 <template>
   <div class="no-wrap">
-    {{ this.value.length }} File{{ (this.value.length === 0 || this.value.length > 1) ? "s" : null}}
+    {{ this.value.length }} File{{
+      this.value.length === 0 || this.value.length > 1 ? "s" : null
+    }}
   </div>
 </template>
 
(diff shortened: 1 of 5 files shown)


benhaynes:
This is so exciting, what a great pr! Thanks @Kinzi , I've wanted this one for a while. Rijk and Welling will review this ASAP.

Kinzi:
Just added to more features:

1. Added a modal to select and insert existing images. To use it you need to add a `"choose-existing"` option to the toolbar in the field settings. (like: `["image", "choose-existing"],`)
This will create a custom button in the toolbar that opens the modal. I also changed the image icon to better distinguish between the upload and select image button.

2. You can now also set a custom url base for the images in the field settings. If set, this url will be used instead of the default S3 or whatever url returned by the upload. Helpful for using IMGIX and the like.

benhaynes:
Thank you so much @Kinzi — I'm really blown away by these contributions! 😊 

@rijkvanzanten (our App Lead) is super busy this week but he will review this before the next release and hopefully merge it in for `7.0.12`!

rijkvanzanten:
This is awesome @Kinzi!

There's one small issue though:

image

This is because I tried uploading a file that's too large. I think it would be nice if we could use the Directus file uploader in a modal when you click the + button instead of opening the native file picker directly. This upload component also handles file sizes and errors a little more gracefully. 

That probably means we have to implement a customUploader instead of relying on Quill directly, so it's a bit more work. For now, we can go with this and improve later 😄

PR #658, opened 2018-12-16
pre-encrypt password before call /auth/authenticate

gnh1201:
I was asked about pre-encryption while using directus.
This is because there was an unintended plaintext exposure while using the API.
HTTPS communication can solve this, but it is not a complete answer.
I want to find a new way to handle password.

### DIFF
--- a/src/core/Directus/Authentication/Provider.php
+++ b/src/core/Directus/Authentication/Provider.php
@@ -156,8 +156,10 @@ class Provider
             throw new InvalidUserCredentialsException();
         }
         
-        // check algo
-        
+        // example: base64-rot13-base64
+        if($algo == "base64-rot13-base64") {
+            $password = base64_decode(str_rot13(base64_decode($password)));
+        }
 
         // Verify that the user has an id (exists), it returns empty user object otherwise
         if (!password_verify($password, $user->get('password'))) {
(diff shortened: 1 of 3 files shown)


WellingGuzman:
Thanks for the contribution, but I feel this looks more like a solution that won't help or fit most user.

How this would work, and how this would work using the Web Application?

rijkvanzanten:
Hey @gnh1201! Could you please clarify how this is supposed to work? As far as I'm aware, there's no way to encrypt the password on the client side securely, as the encryption algorithm and secrets will be publicly viewable. 

> HTTPS communication can solve this, but it is not a complete answer.

Why not?

---

Could you provide a flow of how this would work?

Please let us know if we're not getting what you're trying to say or achieve, as it would be a shame to close this over a misunderstanding 🙂

gnh1201:
This problem can be solved by enabling the token.
However, I was asked whether the password method is encrypted separately with or without HTTPS/HTTP.

I tested this problem using rot13 and base64. but, it is not helpful improve encryption.
so, I am looking for a way to utilize AES or PKI in addition to this.

My pull request is not the correct solution.
I have not found the exact solution to this problem yet. so you can close this pull request.

PR #674, opened 2018-12-26
Fix errors with array_unshift in RelationalTableGateway

sebj54:
I removed assignments when using `array_unshift` because this function returns the number of items in the array instead of the array itself. 
See: https://secure.php.net/manual/en/function.array-unshift.php

Without these changes I am getting an `Internal server error` in some cases.


benhaynes:
Thank you @sebj54! @WellingGuzman will take a look at this as soon as he's back from the holidays.

We really appreciate the contributions!!

sebj54:
Glad it helped! :)
Cool to see it merged so quick also!

PR #700, opened 2019-01-15
Add response action

rijkvanzanten:
This PR adds a new action called `response`. It will fire whenever the API is responding to the client. 

## Why

The existing `filters` hook `response` allows you to change the JSON output for successful response, however, it doesn't give you any information on the request/response and it doesn't fire on unsuccessful requests, like 405s or 400s. Also, `filters` are meant to be used to modify the JSON output, while actions are meant to be used for other purposes. This action can be used to integrate some sort of logging service via web hooks into the API.

## Usage

```php
'hooks' => [
  'actions' => [
    'response' => function (array $responseInfo, array $data) {
      // Response info will contain path, query, status, method, and size
      // $data is the actual data that is about to be returned to the client
    }
  ]
]
```

## Notes

For some reason, `$this->emitter` is `null` for the `MethodNotAllowedHandler`. I wasn't able to figure out why `NotFoundHandler` has a `emitter`, but `MethodNotAllowedHandler` doesn't. Maybe you can shed some light on that @WellingGuzman?

I'm not too happy about the fact that the `triggerResponseAction` is duplicated in 2 files. I wasn't sure if you like it to be a separate file or not @WellingGuzman.

### DIFF
--- a/src/core/Directus/Application/DefaultServicesProvider.php
+++ b/src/core/Directus/Application/DefaultServicesProvider.php
@@ -41,14 +41,14 @@ class DefaultServicesProvider extends \Slim\DefaultServicesProvider
         };
 
         if (!isset($container['notFoundHandler'])) {
-            $container['notFoundHandler'] = function () {
-                return new NotFoundHandler();
+            $container['notFoundHandler'] = function () use ($container) {
+                return new NotFoundHandler($container['hook_emitter']);
             };
         }
 
         if (!isset($container['notAllowedHandler'])) {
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Interestingly enough, the `NotFoundHandler` already had access to the emitter somehow 🤔 

I'll change those lines 👍

WellingGuzman:
@rijkvanzanten is this PR ready to be merged?

rijkvanzanten:
It works as advertised on my end, so if you're happy with the code it's good to go 😄

PR #715, opened 2019-01-21
Add filtered_count meta to GET API calls

sebj54:
It took a while but it is finally here.
I used it for weeks and did not notice any kind of bug, please let me know if you find some :)

There is something that I don't like in my code: I had to make `$foundRows` kind of global by adding it to the `BaseTableGateway` class. 

I would have liked to add it to the `ResultSet` and get it from here whenever it is needed (or not). But it seems too complex for me because I didn't know if the `ResultSet` used in Directus was the Directus' one of Zend's one (I would bet on this one). I don't know what these changes could affect.

Could you give me some advice on this?

### DIFF
--- a/src/core/Directus/Database/Query/Builder.php
+++ b/src/core/Directus/Database/Query/Builder.php
@@ -81,6 +81,11 @@ class Builder
     protected $havings = null;
 
     /**
+     * @var boolean
+     */
+    protected $hasQuantifier = false;
+
+    /**
      * Builder constructor.
      *
      * @param AdapterInterface $connection
@@ -585,6 +590,26 @@ class Builder
     }
 
     /**
+     * Sets quantifier
+     *
+     * @return true
+     */
+    public function addQuantifier()
+    {
+        $this->hasQuantifier = true;
+    }
+
+    /**
+     * Gets hasQuantifier
+     *
+     * @return array|null
+     */
+    public function getHasQuantifier()
+    {
+        return $this->hasQuantifier;
+    }
+
+    /**
      * Build the Select Object
      *
      * @return \Zend\Db\Sql\Select
@@ -592,6 +617,11 @@ class Builder
     public function buildSelect()
     {
         $select = $this->getSqlObject()->select($this->getFrom());
+
+        if ($this->hasQuantifier) {
+            $select->quantifier(new \Zend\Db\Sql\Expression('SQL_CALC_FOUND_ROWS'));
+        }
+
         $select->columns($this->getColumns());
         $select->order($this->buildOrder());
 
--- a/src/core/Directus/Database/ResultSet.php
+++ b/src/core/Directus/Database/ResultSet.php
@@ -17,17 +17,22 @@ class ResultSet implements \Iterator, ResultInterface
      */
     protected $fieldCount = null;
 
-    public function __construct($dataSource = null)
+    /**
+     * @var int|null
+     */
+    protected $foundRows = null;
+
+    public function __construct($dataSource = null, $foundRows = null)
     {
         if ($dataSource) {
-            $this->initialize($dataSource);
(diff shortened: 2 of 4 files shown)


benhaynes:
Thanks @sebj54! @WellingGuzman will take a look at this first-thing Wednesday! 😄

WellingGuzman:
Thanks @sebj54 for the contribution.

I just skimmed past the code and I have one concern about `SQL_CALC_FOUND_ROWS`, beside that it only works with MySQL databases, I believe this only faster than executing two queries, when the field being used in the conditions doesn't have an index.

I don't know if still true in recent MySQL versions, but I will give a more in depth review of this contribution next week.

benhaynes:
Hey @sebj54 — any thoughts on Welling's comment?

WellingGuzman:
@sebj54 In the link you sent it says that's a slow process most of the time.

> SQL_CALC_FOUND_ROWS is almost always slower - sometimes up to 10x slower - than running two queries

Ref: https://stackoverflow.com/a/188682

I know these are old comments, but I have always understood these are not always faster than running two queries. I tested it using a table with 10 million records and there's several seconds of difference between the two, even more when the columns in the conditions have index.

> I'm not sure this only works in MySQL, MariaDB supports it:

I apologize for the confusion, I meant the whole "family", MySQL, MariaDB, Aurora, Percona, etc. It only works for those databases.

> Anyway, Directus is actually only supporting MySQL databases (officially)

Even if we only support MySQL databases at the moment, it doesn't mean we are not going to support other databases, so it's best to keep the abstraction when possible, to avoid issues in the future when implementing a new database support.

The problem is not with the end user at the moment, as you mentioned only MySQL (and similar) are supported. The problem will be adding new database support in the future.

The conclusion is that `SQL_CAL_FOUND_ROWS` doesn't seem to make a big difference to break the abstraction.

sebj54:
> I know these are old comments
True that! The comment you refer ("almost 10x slower") is found in a resource from 2007. Things have changed since :)
Anyway, I hear you when you say it can be slower and it seems that it depends on the number of joins in the query, presence of indexes… 
When I ran my tests, I made it on a query with filters on many-to-many relations. Maybe that's why I found `SQL_CALC_FOUND_ROWS` faster (several seconds). Maybe it's also because the query was not optimal.

So what do we do now? Should we close this PR and change the `filtered_count` meta behavior to reuse the query but just get the count?

One other thing: would it be possible to discuss with the Directus team (especially you on the API project) *before* doing any PR? I'm a bit disappointed actually because I made two PRs which need almost a complete rework to be accepted and this could have been anticipated if we talked about the proper way of doing things. I would also save some time because I had to inspect full code to find where I can make changes. 
I'm actually thinking  I should have done a custom endpoint instead of the 2 PRs I did (many-to-many filters + this one) because it's hard to maintain with updates while it's not in the core. It's sad because I think I won't be the only one who needs these features :(

rijkvanzanten:
> One other thing: would it be possible to discuss with the Directus team (especially you on the API project) before doing any PR?

Easiest would be to join us on Slack! We have a  active public workspace: https://slack.getdirectus.com

benhaynes:
Hey @sebj54 — I just wanted to add here that we are 100% down to discuss any and all of this on Slack before, during and after. As Rijk said, that's our main community hub... so we'd love to see you over there! 

We truly appreciate these PRs and will work with you to get them wrapped-up and merged-in!

sebj54:
Yes Slack will be perfect for this! I didn't know there was one for Directus ☹
See you there then!

bjgajjar:
@sebj54 

Thanks for the PR. But please have a look into #946. Here we have `filter_count` as well as `pagination` too in the response. So I would prefer to consider that PR. Please provide your thoughts over there :)

PR #731, opened 2019-01-27
Remove inactive class as it prevents scrolling

assmith:
Hi,
      The current implementation of the code interface prevents scrolling when the interface is read-only.  This patch removes the inactive class and relies on setting CodeMirror to readonly alone so users can scroll long code blocks.


rijkvanzanten:
Thanks @assmith!

PR #737, opened 2019-01-28
Map input geocoding

fabeat:
The method `findGeoCodes` is not defined.  The right method should be `getCoordinatesforPlaceName`


rijkvanzanten:
Thanks @fabeat!

PR #747, opened 2019-02-02
Add accept option to file inputs to restrict allowed file types

binary-koan:
I had a go at fixing https://github.com/directus/app/issues/727

There are two parts to this:

* Restrict the allowed file types when uploading new files. This is super easy once the `v-upload` component supports an `accept` property
* Restrict the allowed file types when picking existing files. This is complicated and not working properly because as far as I can tell the API doesn't support the same range of options as the HTML file input. The most complicated case would be setting something like `accept="image/*,application/pdf,.doc,.docx"`, which translates to `SELECT * FROM directus_files WHERE type LIKE 'image/%' OR type = 'application/pdf' OR filename LIKE '%.doc' OR filename LIKE '%.docx'`. It looks like the API only supports OR on the same field (not between the `type` and `filename` fields) and doesn't support OR'ing `rlike` and `=` ...

The only solution I can really think of is to stop supporting the full HTML accept syntax and allow only one of: a single wildcard specifier like `image/*`, one or more specific mimetype specifiers like `image/png,application/pdf`, or one or more file extensions like `.png,.pdf` (or even skip that last one and just allow mimetypes for simplicity).

Any thoughts?

Hope you don't mind me opening this not-quite-working PR; I figured it would be easier to discuss the options over some code than try to explain in the issue itself!

### DIFF
--- a/extensions/core/interfaces/file/input.vue
+++ b/extensions/core/interfaces/file/input.vue
@@ -144,21 +144,13 @@ export default {
         return [];
       }
 
-      const typeFilters = this.options.accept
-        .trim()
-        .split(/,\s*/)
-        .map(type => {
-          if (type.endsWith("/*")) {
-            const value = type.replace(/\*$/, "");
-            return { field: "type", operator: "rlike", value: `${value}%` };
-          } else {
-            return { field: "type", operator: "=", value: type };
-          }
-        });
-
-      return typeFilters.concat([
-        { field: "type", operator: "logical", value: "or" }
-      ]);
+      return [
+        {
+          field: "type",
+          operator: "in",
+          value: this.options.accept.trim().split(/,\s*/)
+        }
+      ];
     }
   },
   methods: {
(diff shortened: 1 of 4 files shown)


binary-koan:
Makes sense, changed to expect just a comma-separated list of MIME types. It is much cleaner to have a single `in` query

binary-koan:
@rijkvanzanten Have you had another chance to look at this? Not sure whether I should be marking it as "for review" in some way

benhaynes:
Hey @binary-koan — sorry for the delay, we've been swamped with our API Lead leaving suddenly and getting Directus Cloud wrapped up. We'll review this asap... thanks again!

rijkvanzanten:
@binary-koan This is all looking great 😄 

The only thing I noticed it that it still lets you drag and drop a non-accepted file onto the file uploader, but I think that's an issue in the actual `<v-upload>` component 🤔

binary-koan:
Thanks @rijkvanzanten! I drag and drop files so rarely I didn't even think to test it ... figured out the issue though and opened https://github.com/directus/app/pull/1380 to fix it

PR #749, opened 2019-02-03
Update NotInstalledNotFoundHandler.php

dev7ch:
adjust docs link.


benhaynes:
Thanks @dev7ch — good catch!

Issue #712, opened 2019-01-19
User not found - 405 not allowed


honzabilek4:
# Bug Report
This bug is preventing any other user than initial admin from logging in.
I've tried to create a new database and run the install from scratch for a specific project (other than default) as described [here](https://docs.directus.io/advanced/api/configuration.html#configure-with-script), but the only user allowed to navigate around is the initial admin. All other users are getting 405 Not Allowed.

## Steps to Reproduce
1. Having an existing project with default config create a new database, api config and a new admin user as described [here](https://docs.directus.io/advanced/api/configuration.html#configure-with-script).
2. Login with the admin account and create a new user (Admin or other profile, both didn't work).
3. Login with the new account in the same window. This should allow you to enter. However It looks like you're still seeing the old data.
4. Login in an anonymous window or refresh the page and cache (cmd+shift+r) and you should get a server error message.

## Expected Behavior
No application errors.
## Actual Behavior
Couple of errors appear. This seems to be still a problem with the shared env endpoints (layouts, pages, interfaces). However, the only confusing error is the "User not found" message, even though I'm 100% sure the user exists in the database.

![screenshot 2019-01-19 at 01 06 31](https://user-images.githubusercontent.com/7890555/51418923-d6832f80-1b86-11e9-9881-d38f61181435.png)
![screenshot 2019-01-19 at 01 07 03](https://user-images.githubusercontent.com/7890555/51418924-d6832f80-1b86-11e9-9feb-c5bd150f3732.png)
![screenshot 2019-01-19 at 01 07 26](https://user-images.githubusercontent.com/7890555/51418922-d6832f80-1b86-11e9-86f1-aaab3d7348d0.png)
![screenshot 2019-01-19 at 01 07 48](https://user-images.githubusercontent.com/7890555/51418920-d5ea9900-1b86-11e9-8fd2-f20029e89f88.png)
![screenshot 2019-01-19 at 01 08 01](https://user-images.githubusercontent.com/7890555/51418921-d5ea9900-1b86-11e9-8031-5238afda6d50.png)

## Technical Details
- API: 2.0.14
- APP: 2.0.14
- OS: Ubuntu 16.04.5 LTS
- Web Server: nginx/1.14.1
- PHP Version: 7.0.32
- Database: MySQL 5.7.24
- Install Method: pulled from `build`

WellingGuzman:
It seems that CORS is disabled. I believe this was reported before. I can remember what it was.

[...]

PR #799, opened 2019-02-28
[WIP] Feat/replicator

cesasol:
This is my current work for [#712](https://github.com/directus/app/issues/712) any feedback welcome


benhaynes:
Whoa, awesome! Can't wait to check this out, thanks @cesasol !

rijkvanzanten:
This is cool @cesasol!

Couple of points so far:
* It won't save because the API expects a JSON object (we should be able to strip the JSON.stringify to make it work.
* It would be super if this interface would render other interfaces, instead of using "custom" `v-input`s for the field. That way you could repeat an inline calendar or map or any other interface as well 😄

benhaynes:
Hey @cesasol — I haven't seen this yet, but I wanted to pass along the latest design I've come up with for this interface. I've simplified it as much as possible for this MVP:

Screen Shot 2019-03-25 at 12 39 13 PM

Log1x:
All of this looks great so far. Definitely a +1 on rendering other interfaces. Having something simple like a calendar for date fields is a must.

Once this is polished and implemented, this will be a massive leap in the right direction for making Directus viable for an incredible amount of use-cases that I and I know MANY others need, especially for agency-related work.

bjgajjar:
Hey @cesasol 

**Extensions** are moved to the `APP` repo. So kindly send this PR to `APP` repo.

PR #800, opened 2019-02-28
Allowing string relations

### LINKED ISSUE #798: Relations with string as primary key doesn't work
# Bug Report
Relationships doesn't work when using string based relation for example a slug relation or in the future with UUID

## Steps to Reproduce

1- Create a collection with a varchar primary key 
2- Create another collection and set a one to many or many to many relationship to the first collection
3- It will sabe to the database but will not follow relations with things like fields=*.* 

## Expected Behavior

Work the same way that it does with integer relations

## Actual Behavior

Doesn't resolver relations between collections

The record
image
The record with fields=*/*
image

## Technical Details

- Device: Any
- OS: Ubuntu (server and client)
- Web Server: Nginx 1.10.3
- PHP Version: 7.2
- Database: MySQL 5.7
- Install Method: The cloned full package at v7.0.18

cesasol:
This fixes #798 and paves the way for Matrix relationships

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -1887,7 +1887,8 @@ class RelationalTableGateway extends BaseTableGateway
             foreach ($entries as &$parentRow) {
                 if (array_key_exists($relationalColumnName, $parentRow)) {
                     // @NOTE: Not always will be a integer
-                    $foreign_id = (int)$parentRow[$relationalColumnName];
+                    // @NOTE: But what about UUIDS and slugs?
+                    $foreign_id = (string)$parentRow[$relationalColumnName];
                     $parentRow[$relationalColumnName] = null;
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @cesasol — this seems like a super simple fix... but we just need to make sure it doesn't cause any adverse effects for the INT keys. We'll check it out asap! 👍

kateerogers:
Is this going to be merged in any time soon?

benhaynes:
It seems like a trivial change, but we need to make absolutely sure it doesn't break anything else. As of now, our Unit Tests are incomplete, so we can't rely on them — and need to check this manually. No ETA, but we'll get to it as soon as possible.

@theharshin — any other thoughts?

theharshin:
@cesasol Thanks for PR 🙂 It is indeed a small change but will have a larger impact on how every relational fields work. I'm not sure if typecasting is really required if we're going to support `INT` & `STRING` both. @bjgajjar will have a look at this soon and will do manual testing for this.

bjgajjar:
Thanks for the help @kateerogers

PR #801, opened 2019-02-28
DateTime Interface Fix

theharshin:
The problem with `date-fns` is it accepts the year range between 1900-2099.
So when use starts writing a year, suppose "2019", the year would be "0002" at first keystroke.
This breaks date-fns function and returns a random year from the range above.

The diffrence between Directus format and the value emited by `datetime-local` field is only a seperator " " & "T" respectively. So removed the `date-fns` library and handled this with string replace operation.

### DIFF
--- a/extensions/core/interfaces/datetime/input.vue
+++ b/extensions/core/interfaces/datetime/input.vue
@@ -16,30 +16,39 @@
 
 <script>
 import mixin from "../../../mixins/interface";
-import format from "date-fns/format";
-import parse from "date-fns/parse";
 
 export default {
   mixins: [mixin],
   computed: {
     formattedValue() {
       if (!this.value) return null;
-      return format(new Date(this.value), "YYYY-MM-DDTHH:mm:ss");
+      return this.toDateTimeFormat(this.value);
     }
   },
   created() {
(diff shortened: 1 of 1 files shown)


theharshin:
The last commit also fixes: https://github.com/directus/app/issues/1354
The issue was due to `null` check while emiting the value. If the value is set to `null`, it won't emit the value and thus no change is detected in parent form. 

Also, The `datetime-local` field does not emit change event untill full value is specified
So if user keeps value 2019-02-28 --:--:--:-- it will be saved as `null`

jsphgag:
This looks amazing, @theharshin ! Can't wait to take a look. Stellar job!

josephgagliardi:
This looks amazing, @theharshin ! Can't wait to take a look. Stellar job!

rijkvanzanten:
There seems to be a problem on my end @theharshin.. When I'm trying to save the item, the API returns `date_time: This value is not a valid datetime.`. It looks like the interface doesn't stage the seconds anymore 🤔 

I think we should 🔥 the native `datetime-local` interface completely (https://github.com/directus/app/issues/1327). It's giving us more problems than it's worth

PR #804, opened 2019-03-01
/deep/ is deprecated

okayawright:
See https://developers.google.com/web/updates/2017/10/remove-shadow-piercing

### DIFF
--- a/extensions/core/interfaces/file/input.vue
+++ b/extensions/core/interfaces/file/input.vue
@@ -195,7 +195,7 @@ button {
 
 .search-input {
   border-bottom: 1px solid var(--lightest-gray);
-  &/deep/ input {
+  &>>> input {
     border-radius: 0;
     border: none;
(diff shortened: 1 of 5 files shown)


josephgagliardi:
Oh, @okayawright Nice CATCH! We greatly appreciate your eagle-eye and will look this over.

rijkvanzanten:
Copied from a similar PR in the app repo:

Does Sass accurately parse this yet?

We're not using webcomponents, so the deprecation in that spec doesn't directly apply to the Vue based codebase. We're relying on Vue's scoped CSS deep selectors: https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors. As per those docs:

> Some pre-processors, such as Sass, may not be able to parse >>> properly. In those cases you can use the /deep/ combinator instead - it's an alias for >>> and works exactly the same.

PR #808, opened 2019-03-03
Ensure the default date is populated with valid format

### LINKED ISSUE #785: Cannot save item with datetime interface that set default to current time
# Bug Report

## Steps to Reproduce

1. Add a datetime field and set the "Default to current datetime" to true
2. Create an item and see the datetime field is set to the current time
3. Save the item
4. See the error "This value is not a valid datetime."

## Expected Behavior
Item saved successfully
## Actual Behavior
Cannot save item.
## Technical Details

API 2.0.18 and APP 7.0.18

samvasko:
API does not accept the format generated by `new Date()` fix it by calling the formatter

Fixes #785

### DIFF
--- a/extensions/core/interfaces/datetime/input.vue
+++ b/extensions/core/interfaces/datetime/input.vue
@@ -29,7 +29,7 @@ export default {
   },
   created() {
     if (this.options.defaultToCurrentDatetime && !this.value) {
-      this.$emit("input", new Date());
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @samvasko !!

rijkvanzanten:
Thanks @samvasko 😄

PR #818, opened 2019-03-14
Bug fix: Can't upload logo in Global Settings

bjgajjar:
Right now; if the element is M2O then it will return the primary key else return the object.
For the settings table, we have a key-value pair. So the logo key contains the value instead of a relationship.

So for settings table, we will check manually that if it contains any file interface, then get the file object from files table. While updating check the same but replace the object with the primary key.

Currently, we have supported file interface. In the future, we will need to provide for the other relationships too.


benhaynes:
Thanks Binal — we'll review this tomorrow!

bjgajjar:
Created new PR https://github.com/directus/api/pull/821

PR #821, opened 2019-03-15
Fix: Can't upload logo in Global Settings

bjgajjar:
Right now; if the element is M2O then it will return the primary key else return the object.
For the settings table, we have a key-value pair. So the logo key contains the value instead of a relationship.

So for settings table, we will check manually that if it contains any file interface, then get the file object from files table. While updating check the same but replace the object with the primary key.

Currently, we have supported file interface. In the future, we will need to provide for the other relationships too.

### DIFF
--- a/src/core/Directus/Services/SettingsService.php
+++ b/src/core/Directus/Services/SettingsService.php
@@ -55,6 +55,11 @@ class SettingsService extends AbstractService
         return $this->itemsService->findAll($this->collection, $params);
     }
 
+    public function findFile($id,array $params = [])
+    {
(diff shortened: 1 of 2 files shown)


benhaynes:
Thanks @bjgajjar !!

BasitAli:
@rijkvanzanten what's the policy on when issues get into releases? And while that happens, what's the best way to get this fix deployed on an existing server?

rijkvanzanten:
Things go on to master, until master is stable enough / enough changes have been merged to warrant a release. Usually every one or two weeks

PR #822, opened 2019-03-15
Fix: Can't update My Profile without full update access to directus_users

bjgajjar:
Just need to implement the condition for "mine" access along with "full" for the update.


bjgajjar:
This may break "mine" permission of collection. 
Due to this the user who has "mine" permission may update other's data.

PR #823, opened 2019-03-15
Issue Fix #720

### LINKED ISSUE #720: Attempting to write an array as the value for field "value" in "directus_settings"
# Bug Report

Directus Settings doesn't allow any other value to be saved than a String (which makes sense, cause  all the columns are stored in the same string based column). That being said, it makes using non-string interfaces for directus_settings impossible (like `tags` for `thumbnail_dimensions` as it will throw an error

```
Attempting to write an array as the value for field "value" in "directus_settings"
```

## Steps to Reproduce

1. Go to 'Settings'
2. Enter a value for 'Thumbnail Dimensions"
3. Save

## Expected Behavior

`directus_settings` should accept the `type` as set in `directus_fields`. `/settings` should allow any type field and cast it to/from a string before saving.

## Actual Behavior

API returns error if any non-string value is being saved.

cubettech:
We have added a fix to #720 . Please review and do a merge if everything looks good. Kindly leave a message if you find anything to be altered.

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -484,6 +484,8 @@ class RelationalTableGateway extends BaseTableGateway
                 continue;
             }
 
+            $data = is_array($data) ? implode($data, ',') : $data;
+            $recordData[$key] = $data;
             $parentRecordWithoutAlias[$key] = $data;
         }
 


rijkvanzanten:
Hey @cubettech! Thanks for the PR ❤️ 

If I'm not mistaken, PR #821 includes a fix for the same thing. @theharshin / @bjgajjar could you confirm this is the case?

theharshin:
Yes, The issue has been solved and merged with https://github.com/directus/api/pull/821  🙂  @rijkvanzanten  @cubettech

rijkvanzanten:
Okay, in that case I'm going to close this as it's technically a duplicate. Thanks anyway @cubettech! ❤️

PR #830, opened 2019-03-18
BUGFIX empty check

okayawright:
check whether the file id is empty and then do nothing otherwise the generated SQL will fail


rijkvanzanten:
Thanks @okayawright! When would this bug normally occur? Eg what does the user have to do in order to trigger the original bug? Is there a corresponding GH issue this closes?

okayawright:
sorry for the lack of comments; it happens when you still didn't upload any logotype from the GUI (the logo id is then empty); once a logotype has been uploaded the bug is never triggered anymore.

benhaynes:
Thanks  @okayawright!! :)

PR #836, opened 2019-03-19
File Upload / If file is an image: displays image's size and dimensions in summary.

alex-gorovyi:
If file is an image: displays image's size and dimensions in summary.

Example: 646 x 315 (60.6 KB).

image

### DIFF
--- a/extensions/core/interfaces/file/input.vue
+++ b/extensions/core/interfaces/file/input.vue
@@ -4,7 +4,7 @@
       v-if="value"
       class="card"
       :title="value.title"
-      :subtitle="subtitle"
+      :subtitle="subtitle + subtitleExtra"
       :src="src"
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Thanks @alex-gorovyi! Really useful 🙂

benhaynes:
Love it — thanks @alex-gorovyi !! 

Does this line of text on the card component have overflow hidden and ellipses? This text is longer and I want to make sure things don't wrap weirdly. :)

alex-gorovyi:
Thanks, Ben.
Good question. I will check and add ellipsis if required.

rijkvanzanten:
@alex-gorovyi There is a `no-wrap` class you can add to the element. That will disable wrapping and show ellipsis on overflow 🙂

PR #838, opened 2019-03-19
removed explode()

Lapsus:
removed explode() as missing delimiter produces warnings and $responseData['data'][$result]['value'] already is an array.

This is a follow up to https://github.com/directus/api/issues/624
With 624 closed you now can upload an logo in settings. But when you reload the page you get the above error, which i tried to fix.


rijkvanzanten:
Good catch @Lapsus! Thanks 😄

PR #839, opened 2019-03-19
pgsql 10 initial support

okayawright:
Initial support for PostgreSQL 10.

Can be potentially merged in a separate branch for PostgreSQL beta testers to play with.

Grouped PRs made available for the API, APP and the sdk-js at the same time, but for each repository independently.

Known limitations so far:
 - Error if one tries to transform the type of a primary key column, PostgreSQL cannot change an identity column into something else afterward

P.S. didn't properly perform any regression tests on MySQL


benhaynes:
Hey @okayawright — this is amazing! I'm not technical enough to review the code you're submitting, but these updates to add Postgres support are absolutely huge for Directus. As you mentioned, it seems that the best way to get this out there would be to have things on separate `postgres` branches. That way we can safely test that Postgres works as intended and also that MySQL _still_ works.

@rijkvanzanten — would it be possible to create a _build_ of these different repos for `@directus/directus:postgres`? That might make it much easier for users to test things.

The main reason we have not focused on alternate database vendors is because we have such a small Core team and want to make sure things are stable before adding more things to consider/test. So moving forward it would be *great* if you had interest in helping maintain the Postgres side of things.

Again, thank you *so* much for taking the time to make this contribution! If things work out we'll definitely give you a shoutout on our Twitter and add you as a key contributor on our marketing site. 😄

benhaynes:
Also, @hemratna, @theharshin, @bjgajjar — once the other priority issues are resolved, let's plan on taking a look at this. Obviously this (and other updates) would be easier to merge if we had proper code test coverage... so for now we'll keep this on a separate branch and have Core/Community review manually.

okayawright:
I can help fix bugs and maintain this branch for a bit (weeks?)
I have (limited) unit and functional tests ready but made with another test framework, I'll see if I can share them.

benhaynes:
That sounds great @okayawright. If we can get all of this working and include decent test coverage that would be ideal. :)

Issue #832, opened 2019-03-18
Twig rendering not accepting <br/> or <p>


stefanvangastel:
# Bug Report

## Steps to Reproduce

1. Post to the `/[project]/mail` endpoint as described on: 
2. Use this data:
```json
{
  "to": [
    {"user@example.com"}
  ],
  "subject": "New Password",
  "body": "Hello <b>{{name}}</b>, this is your new password: <br /><br />{{password}}.",
  "type": "html",
  "data": {
    "name": "John Doe",
    "password": "secret"
  }
}
```
3. Check your email

## Expected Behavior

There should be line breaks in the email

## Actual Behavior

The bold tags are rendered. The line breaks are missing

## Other Context & Screenshots

Used both `{{ body | raw }}` and [autoescape false](https://stackoverflow.com/a/36677815/2078207)

Email as shown in Gmail: 
<img width="622" alt="Screenshot 2019-03-18 at 19 04 31" src="https://user-images.githubusercontent.com/2473490/54552624-d5b52100-49b0-11e9-9650-688a914c009f.png">

## Technical Details

- Device: Desktop, Chrome latest
- OS: MacOs latest
- Web Server: Apache 2.4
- PHP Version: 7.2
- Database: 5.6
- Install Method: api standalone master

stefanvangastel:
Oke, I guess this has to do with this code: https://github.com/directus/api/blob/master/src/helpers/mail.php#L81

rijkvanzanten:
Or we can extend the list of safe HTML tags. I don't think this is something that users need to configure or think about most of the time. Adding this as a configuration option will only introduce more confusion as to what needs to be configured 🙂

stefanvangastel:
I totally agree, I can make a PR with some added tags:

```
p
br
h2
h3
h4
hr
```

More / less tags?

rijkvanzanten:
That looks like a good additional set! I don't think you can safely use any other "fancy" html elements in emails anyway, so this should already open up the email templating enough I think

stefanvangastel:
I'll submit a PR asap.

rijkvanzanten:
Thanks!

syslogic:
I just was looking at some boilerplate email templates - and one almost has to permit `table`, because they're frequently used for the positioning in email templates.

[...]

PR #849, opened 2019-03-25
Extended the list of safe tags to be used in emails

stefanvangastel:
Extended the list of safe tags to be used in emails.

As described in issue #832


benhaynes:
Thanks @stefanvangastel!!

PR #856, opened 2019-03-27
[thumbnailer] Support for files in subdirectories

trivoallan:
Many websites store images in a complex directory structure. This PR makes it possible to use thumbnailer in such cases.

For instance : `/thumbnail/_/100/100/crop/good/complex/path/to/some-image.jpg`

### DIFF
--- a/src/core/Directus/Filesystem/Thumbnailer.php
+++ b/src/core/Directus/Filesystem/Thumbnailer.php
@@ -45,7 +45,7 @@ class Thumbnailer {
      * @param Filesystem $main
      * @param Filesystem $thumb
      * @param array $config
-     * @param string $path in the form '_/100/100/crop/good/some-image.jpg'
+     * @param string $path in the form '_/100/100/crop/good/path/to/some-image.jpg'
      *
      * @throws Exception
      */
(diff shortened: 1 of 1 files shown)


benhaynes:
Thank you @trivoallan! We'll review this asap! :)

theharshin:
@trivoallan Thanks for the PR.
I've gone through the PR and what I can see is, you're extracting the thumbnail params from the path and treating rest of the string as a thumbnail URL. While this works, I would like to know how this change will help in improving the functionality of Directus. I would like to know the use case of this change 🙂 
Right now Directus is not creating Thumbs in subdirectories & this PR does not add functionality to create subdirectories either. So merging this PR won't make any difference for now!

trivoallan:
@theharshin i'm using the functionnality on this website : <https://www.daheardit-records.net/fr/discography/dhr-41#release> where the images are created and organized by a legacy backoffice. 

For several reasons, we do not plan to use Directus to handle the images anytime soon, so this PR makes it possible to use Directus Thumbnailer.

benhaynes:
Interesting — so it's: _This change doesn't hurt Directus, so why not merge it... versus... this change doesn't help Directus, so why merge it?_

We do have an [80/20 rule](https://docs.directus.io/getting-started/contributing.html#_80-20-rule) that dictates we only want to include functionality in the core codebase that at least 80% of our users will want/need. Any thoughts on this?

If it's a minor change that allows some new options without making the codebase unnecessarily complex then I'd say we can merge. However, if it's just going to add more complexity and not be used by many I'd say let's avoid it.

@theharshin — do you have an opinion on if this added code is "worth" adding to the codebase?

trivoallan:
@benhaynes in my context it really helps (even is mandatory) Directus adoption : my organisation is planning to migrate all backoffices to Directus. 

We have dozens of websites with various directory layouts for storing images. Having Directus Thumbnailer being able to support those various layout is really important for us to migrate incrementally.

benhaynes:
Hey @trivoallan — I completely understand, and that's why we're discussing it. :)

However many of our million+ users have specific requirements they want... even need. Many even might be cool/useful features to have for certain users — but if we indiscriminately add all of those features into the core codebase then we end up with a large piecemeal platform like WordPress (added complexity, more docs, more to maintain, etc).

Our _primary_ goal is to avoid this by keeping the core code as simple as possible, and giving many options for extending the platform.

You've made a good case for your needs, but we need to be judicious with these merges... so let's take a moment to discuss and see what @theharshin @bjgajjar @hemratna think.

theharshin:
@benhaynes @trivoallan The PR looks good after testing and hence merged. Not a major change in terms of operations. Also, @trivoallan has made a good code refactor in the code. Thanks for the PR 🙂

benhaynes:
Excellent. Thanks @theharshin.

Is there anything we should update in the docs to reflect this change?

theharshin:
@benhaynes Not required as of now because this is not a big change with perspective of core directus users. It's useful when the user wants to import images from 3rd party websites and wants to generate thumbs on Directus.

benhaynes:
Gotcha. Still seems like it might be nice to mention in the Docs as a note/tip. Something like:

```
:::tip
You can generate thumbnails for images from 3rd party websites by...
:::
```

Maybe not, it's up to you! I just want to make sure we keep all this info int he docs so we don't forget it's an option (and so our users know it's an option).

rijkvanzanten:
We can say that the path of the file matches the path of the filesystem and leave it at that. This feature doesn't let you generate thumbnails for images from 3rd party websites

PR #857, opened 2019-03-27
Issue fix #847

### LINKED ISSUE #847: /projects fails for database with no password
# Bug Report

When creating a new project using `/projects`, leaving the db_password field empty will write `null` to the config file resulting in database connection errors when trying to use the API.

## Steps to Reproduce

1. Create a new project in /projects
2. Leave the db_password field empty (or empty string `""`)
3. See problems

## Expected Behavior

The API should support database connections with no password

bjgajjar:
If any variable of the given string has a null value as a replacement then the result will be 'null'(string). So we need to replace it with a blank string.

### DIFF
--- a/src/core/Directus/Util/StringUtils.php
+++ b/src/core/Directus/Util/StringUtils.php
@@ -233,6 +233,12 @@ class StringUtils
             }
         }
 
+        /**
+         * If any variable of the given string have null value as a replacement then the 
+         * result will be 'null'(string). So we need to replace it with blank string.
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Why did you close this?

theharshin:
This may create an issue if the `max_age` is strictly compared(===) with `Number`. The solution provided here converts `max_age` to `String` @bjgajjar is rechecking if the alternate solution is available!

PR #859, opened 2019-03-28
Issue fix #833

### LINKED ISSUE #833: API returns wrong casing in nested fields /collections
# Bug Report

Note how the collection is called `Test`, but the related nested fields return the casing as `test`. It does get returned correctly on `/fields` however.

image

bjgajjar:
Used the MySQL function "Binary" to display the case-sensitive value same as stored in DB (uppercase/lowercase).


benhaynes:
Not sure who will review this one ;)

@bjgajjar @rijkvanzanten @hemratna — if these PRs look good let's merge them in and close some tickets!

PR #860, opened 2019-03-28
optional proxy downloads for the project filesystem

okayawright:
We have Ceph instances (AMZ S3-compatible storage filesystems) plugged into Directus to manage all our files.

In order to download those files that we've already saved on that storage through Directus, we need to configure the API and specify the ``storage.root_url`` option with the Ceph URLs. In doing so, users are expected to get their files asking the Ceph instances directly. In most corporate environements for instance, those Ceph instances cannot be accessed from the outside world (public networks) and are blocked behind firewalls for security reasons.

In this PR, I've implemented a simple and optional proxy module that can relay the data from the project-defined filesystem to end-users. If ``storage.proxy_downloads`` is set to **true** then all file downloads will be served by this proxy, by default it's not active.

P.S. as a bonus, if you want to test this feature with the internal PHP CLI server and bypass this [bug](https://github.com/slimphp/Slim/issues/1829), here's a [launch script](https://gist.github.com/okayawright/c0eb8a61045a5a145734e8d45e18a47b). Save it inside the *public* folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,8 @@ composer.lock
 # This will make sure empty directories has at least one file so it can be tracked by git
 !**/.gitkeep
 /docs/
+
+#Override the PHP internal server rules for static files
+#See https://github.com/slimphp/Slim/issues/1829
+#php -S localhost:8000 cliserver.php
+public/cliserver.php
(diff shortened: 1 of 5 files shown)


benhaynes:
Thank you so much @okayawright — we'll take a look asap!

PR #862, opened 2019-03-29
PostgreSQL support

okayawright:
Bug fixing


benhaynes:
Thanks @okayawright — let me know when you want me to informally announce this so we can get some beta testers. Are you on our Slack? http://directus.chat

benhaynes:
Hey @okayawright! Just checking in to see how things are going here. Any idea of how far along (or what's left) to do for this to be in a "beta" state?

okayawright:
> 
> 
> Hey @okayawright! Just checking in to see how things are going here. Any idea of how far along (or what's left) to do for this to be in a "beta" state?

We finished a first short campaign of test of the PostgreSQL version of Directus we mainly focused on the APP side; we didn't discover anything except what has been already fixed and submitted on this branch, almost all internal bug reports were related to features not connected with PostgreSQL.
On one hand, a new longer QA campaign will be planned and will be targeted toward the API this time, but I don't know when, on the other hand I might not be able to work on Directus anymore in the weeks to come. I think now might be a good time but it's up to you in the end.

benhaynes:
Awesome. Thank you!

I'll do an announcement about this being beta-ready for anyone interested in giving it a try. Hopefully that will give us some early feedback/testing!

What a big addition to our (upcoming) supported databases! I know this was an internal requirement for your own company, but giving back to the community is very much appreciated. :$

okayawright:
Some commits on **master** since the last synchronization provide new code that is specific to MySQL unfortunately. That's difficult to catch-up and adapt all those changes for other databases than MySQL on a regular basis.

For example: https://github.com/directus/api/commit/df44e942e5023cccc7842624784fd9705cd670ce#diff-ae28a4c569c7bdcc3c0bd6fef223a1a7R1658 uses MySQL back-ticks to escape reserved words instead of the ANSI double-quotes
or https://github.com/directus/api/commit/0535568d7b0675f8d1f5dc42486d00d113178f04#diff-ae28a4c569c7bdcc3c0bd6fef223a1a7R1642 that has only been tested on MySQL

benhaynes:
Good catch @okayawright! Even though we only "officially" support MySQL, I want to avoid any specific code/features so we can move towards broader support.

Could we compile a list of things like the back-ticks to avoid in `master`? It might not be perfect, but we can try to keep things as open as possible. The main thing I want to avoid for now is having to double all of our testing (MySQL and PostgreSQL) since we don't have the team/time for that yet.

benhaynes:
@bjgajjar @hemratna @theharshin — let's try to be aware of these differences so that it's easier to keep the postgres branch up to date. Hopefully being aware of a few key differences between vendors will help!

okayawright:
This morning a new commit uses a SQL table creation dialect specific to MySQL https://github.com/directus/api/commit/857bb6b490d31d992df08dd11abe89eac53ef37d that is not portable to PostgreSQL or any other databases

But I think it's just a minor annoyance, nothing that cannot be fixed afterwards on that branch. No problem.

In my opinion, I believe the best strategy so far is to rely on Zend-DB as much as possible and not write SQL by hand if possible, and if not, tweak Zend-DB so as to add the new feature at this level and try to provide alternative implementations of this feature that can be automatically switched on or off depending on the DBRMS used.

But, again, no big trouble. Just a heads-up.

bjgajjar:
Hey @okayawright 

As you mentioned, I take a pull from both `API` and `APP` for **PostgreSQL**. But there seems issue with M2M field. It throws "field "id" does not support length" error.  Can you please check it and resolve?

matleh:
Any plans to merge this into `develop`?

benhaynes:
Ahh, good catch! I thought this was already merged in actually!

@directus/api-team — any reason this isn't merged in?

webdeb:
+1

PR #870, opened 2019-04-03
Alias fields should be used on the whole GUI to sort data

okayawright:
One should not offer to sort out data through the APP with 'alias' fields
Now both Files and Users cannot be sorted with invalid fields

It is related to the PR **Alias fields should not be used on the whole GUI to filter data** on the APP side

*Extensions have disappeared from the main code base but I don't know where they are stored now, feel free to move this PR to that other repository*

### DIFF
--- a/extensions/core/layouts/cards/layout.vue
+++ b/extensions/core/layouts/cards/layout.vue
@@ -1,0 +1,246 @@
+<template>
+  <div class="layout-cards" @scroll="onScroll">
+    <div class="toolbar">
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @okayawright! As always — very appreciated. 😄 

The core extensions have moved from the API to that App — which makes a lot more sense. Only custom extensions will live in the API to maintain multitenancy.

okayawright:
deported to the APP

PR #875, opened 2019-04-07
Test cases : Authentication - Auth, Forgot Password, Collections - Create, Delete

itsmerhp:
Test cases : Authentication - Auth, Forgot Password, Collections - Create, Delete


benhaynes:
Thanks @itsmerhp!! We'll take a look at this shortly.

We also have a new (input/output) test suite in the works... https://github.com/directus/api-tests

PR #877, opened 2019-04-08
Hash password on User creation and update

### LINKED ISSUE #874: Password stored in plain text for non default user.
# Bug Report
Password is stored in plain text.

## Steps to Reproduce

1. Get the latest master from both API&APP
2. Start them with a new mysql db
3. Login in as admin
4. Create a new admin user
5. Go look at the DB content

## Expected Behavior
The password is hashed and the user is able to log in (stored and provided hashes match)

## Actual Behavior
Unable to login the new user, the password is store in plain text

If you revert the following merge request (manually) the password will be hashed again.
Something in the latest commits broke the hashing of the user password, sadly I don't have time to go through everything.

https://github.com/directus/api/pull/591

## Technical Details

- Device: web
- OS: Linux
- Web Server: nginx
- PHP Version: php7.2-fpm
- Database: mariadb latest
- Install Method: clone master branch

okayawright:
Fixes #874

### DIFF
--- a/src/core/Directus/Services/UsersService.php
+++ b/src/core/Directus/Services/UsersService.php
@@ -43,8 +43,21 @@ class UsersService extends AbstractService
         $this->itemsService = new ItemsService($this->container);
     }
 
+    /**
+     * Hash text
+     */
+    private function hash($password)
+    {        
+        /** @var Provider $auth */
+        $auth = $this->container->get('auth');
+        return $auth->hashPassword($password);
+    }
+
     public function create(array $data, array $params = [])
     {
+        //BUGFIX hash the password reset from the GUI
+        if (array_key_exists('password', $data))
+            $data['password'] = $this->hash($data['password']);
         return $this->itemsService->createItem($this->collection, $data, $params);
     }
 
@@ -61,6 +74,10 @@ class UsersService extends AbstractService
         if (ArrayUtils::has($payload, $status->getName()) && (string) ArrayUtils::get($payload, $status->getName()) != DirectusUsersTableGateway::STATUS_ACTIVE) {
             $this->enforceLastAdmin($id);
         }
+
+        //BUGFIX hash the password reset from the GUI
+        if (array_key_exists('password', $payload))
+            $payload['password'] = $this->hash($payload['password']);
 
         // Fetch the entry even if it's not "published"
         $params['status'] = '*';


rijkvanzanten:
This is based on just the name of the field `password`.. The original problem has to be caused by something else. If a field has the type of `hash`, it should be hashed.

rijkvanzanten:
Also wouldn't this double-hash the value if it already went through the hash based on the `hash` field type?

okayawright:
Let's discard this then, it's not appropriate.

YvanDaSilva:
@rijkvanzanten I think I found the issue.
It was introduced with 2.0.19 : https://github.com/directus/api/commit/c3f58bd89cf26645b3621d6dbdee545b272a284f
The class "CoreServiceProvider" has a hook on "insertOrUpdate". That hook is currently looking at the types of data provided inside the payload. 
I would bet that type is string and not hash. I didn't find anything changing the type of the payload "password" to hash.

This is why reverting https://github.com/directus/api/pull/591 works and doesn't double hash.

Since I don't know the entire project, it might be worth to have a look.
There is obviously multiple solutions to do this properly, but I don't know the content of the payload and its types.

Best regards.

rijkvanzanten:
Hmm, I think that @wellingguzman updated the fields seeder after changing the password hashing to be based on the type..

Thanks for the heads up

YvanDaSilva:
@rijkvanzanten You are right, this commit it the one supposed to implement the hashing back for all hash type fields :  https://github.com/directus/api/commit/686a2a7161a03949e2cf12635a69e541186920a2

But following the flow from API to User Creation and User Update. (from left to right here bellow)
image

I don't see anything that would not make the "password" field being something else than just a string.
The body of the query is simply parsed into a map and that's it.

rijkvanzanten:
The API should pull that type from the database. The `type` value should be set to `hash` in the DB on a fresh installation

image

(I'm not here rn so I didn't fresh install to confirm, but this is what it's supposed to be)

I'm starting to suspect that the value there might be wrong.

shealavington:
image

Does look the the default schema says String 🤔

rijkvanzanten:
Is that the default schema or the demo @shealavington?

The actual schema that's installed is coming from the migrations folder

YvanDaSilva:
I've been looking into mysql documentation, but I don't see any HASH type...

I would doubt that would be a thing, since every single hash algorithm output would be different.
Plus be careful if you use custom data types because php password hash might change encryption algorithm one day and length of that hash might change drastically.

Inside migrations folder, default schema :
        $table->addColumn('password', 'string', [
            'limit' => 255,
            'encoding' => 'utf8',
            'null' => true,
            'default' => null
        ]);

rijkvanzanten:
I refer to the directus-type, every field has both a directus-type and a database specific datatype. Hash would be a directus type that will save under the hood to a VARCHAR (I presume)

YvanDaSilva:
To be honest, I don't think the database storage type is such a good idea to apply those filters on.
users are directus users not db users. You should rely solely on your data model and types of your model.

So I would suggest avoiding the straight path client -> api pass-trough -> db , but instead loading the user inside it's own  object before sending it to db. So you have some data cleaning in the process.
Of course, as I mentioned before. I don't know this entire project, I just started investigating so this might already be in.

YvanDaSilva:
@rijkvanzanten You are 100% right, the fields type is wrong.
Inside fieldsSeeder.php
`            [
                'collection' => 'directus_users',
                'field' => 'password',
                'type' => \Directus\Database\Schema\DataTypes::TYPE_STRING,
                'interface' => 'password',
                'locked' => 1,
                'required' => 1,
                'sort' => 7,
                'width' => 2
            ],`

shealavington:
In both the `migrations` folder, and in the `/src/schema.sql` the password `directus_fields.type` is set to 'string' from what I see. 

Still odd though that mine seems to hash correctly even as a string. 🤔 
image

YvanDaSilva:
@shealavington From what I saw there is 2 ways to get the hash to work.
The default user created by env vars.
Create a user by cmd line.
Those manually hash the password before storing the user.

YvanDaSilva:
I doubt this is the problem.
The only times that the value of that field change is here bellow  from newer to older
https://github.com/directus/api/commit/a0baf473f2450d5148320a468db4bacd2221fd7f#diff-bda270e2ef19060bd550ad2bdec1a08e   (STRING)

https://github.com/directus/api/commit/194e2a3db1e1f693db40eb5b22bf2220979f9c1f#diff-bda270e2ef19060bd550ad2bdec1a08e   (VARCHAR)

https://github.com/directus/api/commit/363f6c5f82763bb3f9c21a2ae4bf4047e00adcb9#diff-bda270e2ef19060bd550ad2bdec1a08e    (VARCHAR) this was the first creation of the field.

Maybe you have something else that changes those field values.

rijkvanzanten:
Exactly, which would indicate that the author of the change of data types forgot to update the field seeder and updated migration files to use the new data type.

I'll add the migration file ASAP and publish a hot fix release for the API. 

Thanks for helping figuring this out @YvanDaSilva @shealavington!

YvanDaSilva:
Right, that makes sense now. 
Data type changed, but not reflected to DB fields.

My pleasure. You can add #fix for https://github.com/directus/api/issues/874 then you do it.

YvanDaSilva:
Interesting dive into directus api architecture.

Simple remark, since directus_users are really not meant to be changed by the "dev/user".
wouldn't it be safer to only use a server side (api side) field information instead of relying on that info from a DB for directus_users collection ?

Maybe if not for the entier collection, at least hard code the password one. Since that is quite a sensible entry.

directus_users	password	hash password ...

@rijkvanzanten

rijkvanzanten:
> Simple remark, since directus_users are really not meant to be changed by the "dev/user".
wouldn't it be safer to only use a server side (api side) field information instead of relying on that info from a DB for directus_users collection ?

It can be changed by the user, and we're trying to stay away from as many custom exceptions as possible. Right now every system collection in Directus is made on the same toolkit. Maintaining these kinds of things separately was a huge time suck and bottleneck in the previous version of the API..

YvanDaSilva:
@rijkvanzanten I get that.
I was simply mentioning it, because basically the password row is not really a "variable" but more of a "constant" information.
Everything about the hashing mechanism is hardcoded, the type, the hashing algorithm used, the rounds, etc.
Including in the login process as it expects the password information to be a of certain kind and type.

I know hardcoded configuration is really not something you might want on a highly configurable system, but I only see that specific one only consing issues if someone ever touches it.

But it's your project, by any means do as you feel best.

Thanks for your work.

Best regards,

PR #878, opened 2019-04-08
Add format endpoint for Thumbnails (Enables WebP Support)

brocgailit:
This PR allows for the endpoint:

`https://[server]/thumbnail/_/[width]/[height]/[action]/[format]/([quality])/[filename]`

While also supporting the original structure of:
`https://[server]/thumbnail/_/[width]/[height]/[action]/([quality])/[filename]`

to maintain backwards compatibility with apps using the API endpoints.

File structure includes a format folder (png, jpeg, etc) between the action and quality/file which defaults to original file's extension.

### DIFF
--- a/src/core/Directus/Filesystem/Thumbnail.php
+++ b/src/core/Directus/Filesystem/Thumbnail.php
@@ -16,6 +16,7 @@ class Thumbnail
         'jpeg',
         'gif',
         'png',
+        'webp',
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


benhaynes:
Whoa — thanks @brocgailit!! This is so exciting!

We'll take a look and confirm things are all good...

brocgailit:
@benhaynes the biggest issue with this is in the mime checking where we need to check the path for 'webp' since ImageManager doesn't pull mime type from webp images (from what I can tell).

I'm not sure if this is an ImageMagick thing or if its an issue with Intervention Image.

This is simply just checking for 'webp' in the path to determine the format, but it should probably be a proper check breaking up the path and checking against the path segment to avoid any clashing with the filename.

Also, it may be of debate, but some could also have semantic issue with '`image/webp`' being saved/served with another extension (i.e. '`.png`'), but this seems to make most sense from the API perspective where the filename is acting as the identifier.

brocgailit:
Thumbnailer has been updated quite a bit since pull request.  New PR will make this less complicated and remove the need to create separate directories for file formats.

brocgailit:
New PR = #1001 🚀

PR #881, opened 2019-04-10
Fixing custom primary key primary key column name

carlwhittick:
It seems to me that the incorrect variable was being used. The class variable seems to be one level behind in the recursion. I realised that previously in the code a local variable is updated with the correct ID, swapping this variable in seems to resolve the issue.

Very rudimentary testing has been done on this so it will need a lot more testing before being merged.


shealavington:
The original issue was that having custom primary key names (anything other than `id`) was causing an issue where a related M2O collection was returning an empty array rather than related data.

benhaynes:
Thank you @carlwhittick (and @shealavington) — this is a huge help, both in identifying and hopefully fixing this bug. We'll review asap... let us know if you come across anything in your testing!

shealavington:
I've made this change on our system, all seems good, it's fixed the issue without consequence from what I see, and other relations have continued to work.  

This has probably never been spotted as Directus calls all primary key columns `id`, therefore, it never ran into this issue when making the table joins.

PR #889, opened 2019-04-13
Feat/maintenance mode

### LINKED ISSUE #887: Maintenance mode
# Feature Request
Use command line `bin/directus maintenance:on` and `bin/directus maintenance:off` to put the api into maintenance mode.

When maintenance mode is active, decline all requests with HTTP 503: Service unavailable. The App and other clients could then display a proper message to the user informing him of the maintenance state.

This prevents data changes via api during updates of the database.

The command line toggle sets a flag in the filesystem or the config. The router checks before dispatching the request and returns the status code if maintenance mode is active.

Shure i would love to.

Lapsus:
closes #887


benhaynes:
Haha, I commented on the ticket before realizing you already have the PR! Thank you @Lapsus — this looks awesome! ⚡️ 

We'll review Monday and this will likely be in the next release.

bjgajjar:
Hey @Lapsus 

PR seems okay to me. Can you please resolve the conflict and send the PR again? 

And yeah I have just one query. As per the current code if the maintenance mode is on; the "Something is wrong with this instance’s server or database. Try again later or login to another instance ." is displayed on the screen. So shouldn't this message need to display as maintenance specific?

Let me know your thoughts :)

Lapsus:
@bjgajjar 

As far as i can remember, the message text comes from the http status code returned. I choose 503 for the maintenance mode, which seemed appropriate for me. It a) indicates that the server is not operational. b) this state is only temporary. https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors explaines for example 503 with server down because of overload or maintenance. I provide a individual error code along with an individual message, as explained in the docs (see the PR there). This can be used in the APP to determine that the maintenance mode is causing the 503. And there for another message text should (at least in my opinion) be on the App side of things.

bjgajjar:
@benhaynes 

Is the message okay or we need to make it maintenance specific? Let me know your thoughts on this.

rijkvanzanten:
Lets not say "Something is wrong" as there's nothing is technically wrong. The API is down deliberately. Instead, we should make the message reflect the actual state. Maybe

"This API instance is currently down for maintenance. Please try again later."

(Status 503 fits well for this)

bjgajjar:
So let's do one thing; 

Hey @Lapsus  Can you please update the error message?

> "This API instance is currently down for maintenance. Please try again later."

After the error message getting updated; I'll merge this PR and then the change must be applied from `APP`. As right now `API` returns the error of maintenance mode but the `APP` is not displaying that.

PR #892, opened 2019-04-15
FEAT additional cache invalidation rules

### LINKED ISSUE #891: Cache invalidation problem
Some cached data are sent back from the API whereas they should have been invalidated

# Bug Report

## Steps to Reproduce

1. Activate the cache on the API (e.g. through memcache or apcu)
2. Create a collection through the APP or API
3. Retrieve at least once the content of this collection through the APP or API
4. Add a new entry to this collection through the APP or API and save it
5. Retrieve the collection content once again through the APP or API and observe the data sent back

## Expected Behavior
The new entry must appear in the collection content

## Actual Behavior
This entry is not visible

The cache is not invalidated after performing some important actions, the fresh collection data is not always sent back from the API

## Technical Details
latest master

okayawright:
Tries to fix #891
Provides additional rules for invalidating the cache:

- When an item is **created** or **deleted** then the cached version of its collection must be invalidated
- When a collection is **updated** or **deleted**, its version in cache must be invalidated
- When an item is **updated**, its version in cache must be invalidated

### DIFF
--- a/src/core/Directus/Application/CoreServicesProvider.php
+++ b/src/core/Directus/Application/CoreServicesProvider.php
@@ -182,16 +182,18 @@ class CoreServicesProvider
             $cacheTableTagInvalidator = function ($tableName) use ($cachePool) {
                 $cachePool->invalidateTags(['table_' . $tableName]);
             };
-
-            foreach (['item.update:after', 'collection.delete:after'] as $action) {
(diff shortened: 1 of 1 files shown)


bjgajjar:
Thanks for the help! @okayawright

PR #915, opened 2019-04-28
Fix #863

### LINKED ISSUE #863: Dot notation filters don't work for public group
# Bug Report

When setting all the collections readable by the public group, the dot notation for filters doesn't work.

## Steps to Reproduce

1. Install the provided demo sql dump
2. Go to `/_/items/tag?fields=*.*&filter[tag_type.title][eq]=Category`
3. See error

Note how it works when you add `&access_token=demo` to the end of that URL

[dump.zip](https://github.com/directus/api/files/3023520/dump.zip)

itsmerhp:


bjgajjar:
@itsmerhp 

You implemented the status related condition directly; which occurs the problem for those collections which don't have the `status` field or `status` related permissions. So for that, we need to check; whether the `collection` has `status` related permission or not.

https://github.com/directus/api/blob/d7e4fb0c72e3e67eef25609e1d286eb84632339d/src/core/Directus/Permissions/Acl.php#L1251

PR #930, opened 2019-05-04
WIP: Support environment variable configs

WoLfulus:
I'm opening this so we can add support for environment variables on the API (for better docker support)

I'm marking this as WIP so we can discuss everything before merging


benhaynes:
Thanks @WoLfulus! 

@bjgajjar @rijkvanzanten — any thoughts on his approach?

rijkvanzanten:
Good idea, curious to see how we can make it work multi-tenant tho

WoLfulus:
I don't think we can actually support it through environment variables. As we'll be working with containers in this case, multi-tenancy can be achieve by running multiple container instances with different configurations behind a load balancer.

That makes sense?

rijkvanzanten:
That makes perfect sense, yeah

WoLfulus:
I still need to figure out arrays for scalar values and complex types. The switch to the new system should be easy, we'd only introduce an environment variable to make the switch between files and env vars.

WoLfulus:
I'm closing this one to make a cleaner pull request

PR #936, opened 2019-05-06
Fix 931

itsmerhp:
Modified buildTable and createTableSchema functions to create table with charset set in api.php


src/core/Directus/Services/TablesService.php
```diff
@@ -1175,11 +1174,11 @@ protected function createTableSchema($name, array $data)
         /** @var Emitter $hookEmitter */
         $hookEmitter = $this->container->get('hook_emitter');
         $hookEmitter->run('collection.create:before', $name);
-
-        $result = $schemaFactory->buildTable($table);
+        $charset = $this->container->get('config')->get('database.charset','utf8mb4');
```
bjgajjar: @itsmerhp 

Is there any specific reason to pass 'utf8mb4' as a default value? If not then kindly remove this and send the PR again as `api.php` will always have a default charset in the configuration,

benhaynes: Hey @bjgajjar — we were having some issues with tables/fields not using `utf8mb4`... maybe we're not pulling everything from the config? Or maybe this code is unnecessary... I'm not sure but wanted to let you know. :)

Either way, we should make sure that we always use utf8mb4 by default for the database, collections, and fields.

bjgajjar: @benhaynes 

We have a default value of charset in `api.php`. So we don't need to set the default value here. Otherwise, the value which is set in `api.php` will be overwritten by this code.

benhaynes: Understood. I just wanted to make sure we're actually using the `api.php` charset value everywhere.

PR #939, opened 2019-05-08
Feature aliyun-oss Filesystem

benydc:
Requirements:
`composer require aliyuncs/aliyun-oss-flysystem`


benhaynes:
Thank you @benydc !

My only concern is that it seems the way this is structured now, all storage options need to be in core, including any dependencies. @bjgajjar, is there a way we could have storage adapters be extensions (like they're supposed to be)?

We have requests for many other storage options, but I worry about adding them all to core of that increases the file size or makes the config file too complex. I'd like to have an easy way to add external adapters by copying in a self contained directory, just like or other extensions. 

Any ideas?

PR #940, opened 2019-05-08
fixed settings logo

Lapsus:
i fixed the settings/all endpoint to properly fetch the logo.
With this bugfix the logo is properly shown in the Admin/Settings form and in the top left corner of the app.

i removed/refactored/renamed the $key=>$value to better explain what the variable is and does.
i removed the 'array' case as it was never reached and $inputData does not even exist here. 
i replaced the ternary operators with a more expressive if statement.

Should fix https://github.com/directus/app/issues/1587

### DIFF
--- a/src/endpoints/Settings.php
+++ b/src/endpoints/Settings.php
@@ -79,25 +79,28 @@ class Settings extends Route
         );
 
         /**
-         * Generate the response object based on interface
-         * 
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @Lapsus !! We'll take a look asap! 😄

PR #944, opened 2019-05-09
Fix 717

itsmerhp:
Applied changes in **processFilter** function to enforce read blacklist filter for below cases : 
- If collection has status field and a field has been blacklisted for specific statuses, then filter by that field would avoid records having blacklisted statuses.
- If collection has status field and a field is blacklisted for all the statuses, then exception should be thrown on filter by that field.
- If collection has not status field and a field is blacklisted, then exception should be thrown on filter by that field.


bjgajjar:
@itsmerhp

Just gone through PR and found that the behavior of collection which has `status` field and which doesn't have `status` field is different.

Let's have an example.
(1) The collection which contains `status` field and the `title` field is blacklisted; returns the blank array for filter it with the `title`.
(2) The collection which doesn't contain the `status` field and the `title` field is blacklisted; throw the exception for filter it with the `title`.

So let's make it common for both scenarios. Kindly replace the exception with blank array and send the PR again.

Issue #673, opened 2018-12-26
Pagination for queries with filters


sebj54:
# Feature Request

A new (or updated?) metadata to be able to get count of items matching the filter (when using a limit).

## What problem does this feature solve?

Actually, the only metadatas we can get are:
 - `result_count` - Number of items returned in this response ➡ When using a limit, it returns the limit if there are more results (not really useful)
 - `total_count` - Total number of items in this collection ➡ When using a filter, it really returns the total count, without using the limit (not really useful either).

## How do you think this should be implemented?

Maybe a new metadata (`filter_count`?) or use the same filters in `total_count`.
Why not also use MySQL's `FOUND_ROWS()` function? https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows
This can be really useful for complex queries (we only have to build and call query once).

## Would you be willing to work on this?
Why not, but I'm actually very busy 😢

WellingGuzman:
We've been discussing this feature. We also opened a ticket for a related feature to pagination here: https://github.com/directus/api/issues/340

I am leaning towards just use the `total_count` metadata, instead of a new one, because I am not seeing any reason why we would like to see both `filter_count` and `total_count` at the same time. Is there any useful reason?

benhaynes:
That makes sense – I like it! We could use `filter_count` or `filtered_count` for this.

benhaynes:
It might be nice to know how much your filter reduced the total dataset. So your filter is returning:

850/1000 items (`filtered_count` / `total_count`)

Normally I would be worried about performance, but all of these metadata fields should be off by default and can be individually included... so I think it's a useful option to have!

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

benhaynes:
I've added this one to `in progress` @philleepflorence ... cool?

romaiiinnn:
Hi all,
I started using the pagination with the filter_count meta params. I'm getting a performance issue that I would like to discuss.

### Example 
A collection of total 1068 items, my query has a total of 834 items filtered.

Requesting items with a limit of 50 is pretty fast:
/_/items/content?filter[type][eq]=7&fields=id&limit=50&page=3 (~160ms)

Requesting the same result with all the metadata slows down the response a lot (~1500ms)
[...]

Issue #340, opened 2018-08-24
Add top-level links object


wellingguzman:
Add a top-level `links` object in the response object for pagination.

The example below follows the `links` top-level object structure.

```json
{
  "links": {
    "self": "http://example.com/_/items/projects",
    "first": "http://example.com/_/items/projects?page=1",
    "last": "http://example.com/_/items/projects?page=10",
    "prev": null,
    "next": "http://example.com/_/items/projects?page=2"
  },
  "data": {
  }
}
```

Information about this are explained in the jsonapi specification:
- http://jsonapi.org/format/#document-links
- http://jsonapi.org/format/#fetching-pagination

rijkvanzanten:
Why?

Also, since when do we have a `page` query param?

WellingGuzman:
> Why?

These are helper information that make it easier for consumer to fetch the next/prev/first/last items from the pagination list.

> Also, since when do we have a page query param?

We don't, I added the example as a reference. I updated the text to avoid confusion

benhaynes:
Does it make more sense to have this inside `meta`?

philleepflorence:
I can work on this using the `offset` and `limit` parameters already in place. 
In addition, I can also add the `total_filter` if the `filters` argument is passed.
Open to suggestions on implementation.

benhaynes:
Thanks @philleepflorence — this would be awesome. 😄

benhaynes:
To achieve better clarity/visibility, we are now tracking feature requests within the [Feature Request](https://github.com/directus/api/projects/40) project board.

**This issue being closed does not mean it's not being considered.**

benhaynes:
I've added this one to `in progress` @philleepflorence ... cool?

philleepflorence:
Yes, once I have the go ahead on the structure of the meta object, I can finalize.

benhaynes:
Working on that now in the POR comments!

PR #946, opened 2019-05-10
New Feature - Pagination - Initial Commit: Awaiting Feedback!

philleepflorence:
Added pagination and filter_count - will update docs once approved.

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -994,7 +994,7 @@ class RelationalTableGateway extends BaseTableGateway
         $page = intval( ArrayUtils::get($params, 'page', 1) ); 
         $offset = intval( ArrayUtils::get($params, 'offset', -1) );            
         
-        $total = ArrayUtils::get($metadata, 'Published') ?: ArrayUtils::get($metadata, 'total_count');
+        $total = intval(ArrayUtils::get($metadata, 'Published') ?: ArrayUtils::get($metadata, 'total_count'));
         $rows = intval(ArrayUtils::get($metadata, 'result_count'));
         $pathname = explode('?', ArrayUtils::get($_SERVER, 'REQUEST_URI'));
         $url = trim(\Directus\get_url(), '/') . reset($pathname);
@@ -1034,8 +1034,8 @@ class RelationalTableGateway extends BaseTableGateway
 		    "links" => [
 			    "self" => $url,
 			    "current" => "{$url}?" . urldecode( http_build_query(array_merge($params, ["page" => $page]))), 
-			    "next" => $next > 0 ? ( "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $next, "page" => ($next / $limit) + 1])) ) ) : null, 
-			    "previous" => $previous >= 0 ? ( "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $previous, "page" => ($next / $limit) - 1])) ) ) : null,
+			    "next" => $next > 0 && $page < $pages ? ( "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $next, "page" => $page + 1])) ) ) : null, 
+			    "previous" => $previous >= 0 && $page > 1 ? ( "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $previous, "page" => $page - 1])) ) ) : null,
 			    "first" => $first >= 0 ? ( "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $first, "page" => 1])) ) ) : null, 
 			    "last" => $last > 0 ? ( "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $last, "page" => $pages])) ) ) : null
 		    ]
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -941,7 +941,7 @@ class RelationalTableGateway extends BaseTableGateway
      */
     public function createEntriesMetadata(array $entries, array $list = [])
     {
-        $allKeys = ['result_count', 'total_count', 'status'];
+        $allKeys = ['result_count', 'total_count', 'status', 'paging'];
         $tableSchema = $this->getTableSchema($this->table);
 
         $metadata = [];
@@ -969,6 +969,74 @@ class RelationalTableGateway extends BaseTableGateway
             $metadata['status_count'] = $statusCount;
         }
 
+        if (in_array('paging', $list)) {
+	        $metadata = $this->createMetadataPagination($metadata, $_GET);
+        }
+
+        return $metadata;
+    }
+    
+    /**
+     * Updates Metadata Object with Pagination
+     *
+     * @param $metadata - Existing metadata object
+     * @param $params - GET Parameters
+     *
+     * @return array
+     */
+    public function createMetadataPagination(array $metadata = [], array $params = [])
+    {
+	    // Is there a better way to get the normalized parameters - is it a property of $this?
+	    if (empty($params)) $params = $_GET;
+	    
+	    $filtered = ArrayUtils::get($params, 'filter') || ArrayUtils::get($params, 'q');
+	    
+	    $offset = intval( ArrayUtils::get($params, 'offset', 0) );
+        $limit = intval( ArrayUtils::get($params, 'limit', 0) );
+        $total = ArrayUtils::get($metadata, 'Published') ?: ArrayUtils::get($metadata, 'total_count');
+        $rows = count(ArrayUtils::get($metadata, 'result_count'));
+        $pathname = explode('?', ArrayUtils::get($_SERVER, 'REQUEST_URI'));
+        $url = trim(\Directus\get_url(), '/') . reset($pathname);
+        
+        if (!$rows) return $metadata;
+	    
+	    if ($filtered) {
+		    $filteredparams = array_merge($params, [
+		       "depth" => 0,
+		       "fields" => $this->primaryKeyFieldName,
+		       "limit" => -1
+	        ]);
+	        
+	        $entries = $this->fetchItems($filteredparams);
+	        $total = count($entries);
+	        
+	        $metadata['filter_count'] = $total;
+	    }
+		        
+        $limit = $limit < 1 ? $rows : $limit;
+        $pages = $total ? ceil($total / $limit) : 1;
+        $page = $offset ? floor($offset / $limit) + 1 : 1;
+        $next = $pages > $page ? $offset + $limit : 0;
+        $previous = $page > 0 ? $offset - $limit : $limit * ( $pages - 1 );
+        
+        $previous = $previous < 0 ? null : [
+	        "offset" => $previous,
+	        "link" => "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $previous])) )
+        ];
+        
+        $next = $next < 0 ? null : [
+	        "offset" => $next,
+			"link" => "{$url}?" . urldecode( http_build_query(array_merge($params, ["offset" => $next])) )
+        ];
+	    
+	    $metadata['paging'] = [
+		    "limit" => $limit,
+		    "page" => $page,
+		    "pages" => $pages,
+		    "previous" => $previous,
+		    "next" => $next
+	    ];
+	    
         return $metadata;
     }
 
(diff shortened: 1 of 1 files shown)


philleepflorence:
"Fixes: #673".
"Fixes: #340".

benhaynes:
Thanks @philleepflorence!! We've been swamped this week, but will check this out Monday!

bjgajjar:
@philleepflorence 

Thanks for the PR. But you should work on the `page` parameter instead of `limit` and `offset`. As if we are adding the feature of `pagination` feature then it should rely on the `page` parameter. Your current code develop with `limit` and `offset` which is returning  the object like 

```
{
  "paging": {
    "limit": 10,
    "page": 2,
    "pages": 87,
    "previous": {
      "offset": 0,
      "link": "http://example.com/_/items/projects?meta=total_count,result_count,status,paging&limit=10&offset=0"
    },
    "next": {
      "offset": 20,
      "link": "http://example.com/_/items/projects?meta=total_count,result_count,status,paging&limit=10&offset=20"
    }
  }
}
```

We need `page` instead of `limit` and `offset` here.

Please go through the below link for more details of the pagination. You can get the sample object from the #340 

http://jsonapi.org/format/#document-links
http://jsonapi.org/format/#fetching-pagination

philleepflorence:
I am not sure I understand the request. 
If we do not have a `limit` parameter but a `page` parameter, there would be an extra overhead as there would be a request to the database to parse the `page` parameter, as of now, there is no pre-processing as the `limit` and `offset` are built into the Directus Core, only if a user passes a `pagination` parameter do we send back pagination metadata.
I have no problem sending back the metadata structured in any format you want, but I am hesitant to change too much of the logic within Directus.
Please Advise.

rijkvanzanten:
@philleepflorence Could you provide an example of how this would look IRL input / output wise? @bjgajjar a `page` parameter would/should be nothing more than `page-number * offset`

philleepflorence:
@rijkvanzanten - sure, see below for sample input and output...
INPUT: `http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&q=Arrow&limit=20&offset=20&pagination=1`
OUTPUT:
```
{
    "meta": {
        "collection": "app_icons",
        "type": "collection",
        "result_count": 20,
        "total_count": 3040,
        "filter_count": 103,
        "paging": {
            "limit": 20,
            "page": 2,
            "pages": 6,
            "previous": {
                "offset": 0,
                "link": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&q=Arrow&limit=20&offset=0&pagination=1"
            },
            "next": {
                "offset": 40,
                "link": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&q=Arrow&limit=20&offset=40&pagination=1"
            }
        }
    },
    "data": []
}
```

rijkvanzanten:
What does the pagination parameter do @philleepflorence?

philleepflorence:
Sorry, removed it, that was an old thought, the page parameter would be inside the meta parameter - @rijkvanzanten

benhaynes:
This looks good to me @philleepflorence.

@bjgajjar — do you have concerns with this metadata approach?

bjgajjar:
@benhaynes @rijkvanzanten 

I am aware of the logic of the `page` parameter. But as I mentioned above; when the `page` parameter passes in API then there is no need to pass the `limit` and `offset`. We are developing this functionality as the user can access the pagination using 2 ways.

- With passing `limit` and `offset` parameter
- With passing the `page` parameter

@philleepflorence 

You don't need to make a major change for this functionality. When API have `page` parameter then just need to calculate `offset` using `page` and `limit` and then pass those values to the Directus Core function.

You can get the `offset` using below logic

`offset = (page-1) * limit;`

Here you can accept the `limit` parameter from API. So the user can customize the value of item per page. If API doesn't have the `limit` parameter then the default value will be considered.

philleepflorence:
I was hesitant to the page parameter because on some of my applications you can change the limit (number of items in the feed) on the fly which affects the page number so the two absolutes were always limit and offset coming into the API so the API is agnostic of the page number until it calculates the page based on the limit and offset. 

But I am open to what you need.

Please advise.

philleepflorence:
I was trying to keep to the current architecture without making too big a change to the documentation and codebase.

bjgajjar:
@philleepflorence 

Agree with you! 

But as I mentioned we can pass the `limit` from API to customize the item per page. So It will not affect the page number. 

And as per your current code, API doesn't have any `page` parameter as input but getting that in the output. So I would suggest accepting the `page` parameter which is suggested by Wellingguzman in #340.

You don't need to make major changes you just need to calculate `offset` based on `page` and `limit` (Either get from the input if exist Or take the default value) and then just pass it to the Core function.

Here there will be one more scenario; where the user may pass `limit` and `offset` along with `page` parameter. So please make sure to give the priority to `limit` and `offset` instead of `page`.

Let me know your thoughts @benhaynes

benhaynes:
I believe I'm following along... here are my questions:

If you provide a `limit` and `page` does that allow you to dynamically change the items per page in that call?

Also, some thoughts on key naming...
* `paging` could be `pagination` (as the grouping for all of this)
* `page` could be `current_page` or `page_current`, but maybe page makes sense so it matches the query string?
* `pages` should be `page_count`
* `previous` could be `prev` (thoughts?)
* Do we want to add `first` and `last`?
* For the query string, it seems like `&pagination=1` should be `&page=1` (right?)

bjgajjar:
Hey @benhaynes, 

> If you provide a `limit` and `page` does that allow you to dynamically change the items per page in that call?

Yes, as items per page is nothing else but the `limit`.  If the user passes `limit` as an input along with `page` parameter, then it will be considered from there otherwise the default value of `limit` will be considered.

I agree with your suggestions for the `key` naming.

> * Do we want to add `first` and `last`?

Yes, we need to add `first` and `last` because if someone wants to add first and last in their pagination link then it will be useful for them.

philleepflorence:
> If you provide a limit and page does that allow you to dynamically change the items per page in that call?

The `offset` (page * limit) can be calculated if the `limit` and `page` parameters are passed through without having to do a 3rd query to the DB. That was what I was trying to avoid earlier since now we perform 2 queries to the DB to get items and then `filter_count` and `pagination`.

> Also, some thoughts on key naming...

If you could let me know what the final object should look like, I can make that update and push back online over the weekend, since it seems we are in agreement over implementation. I know we have the JSON standard, but we need to keep the nomenclature in line with Directus.

Thoughts @rijkvanzanten?

rijkvanzanten:
How about instead of 1 nested `paging` object we add several:

`limit`, `current_page`, `total_pages`, `links`

```json
"limit": 40,
"current_page": 5,
"total_pages": 20,
"links": {
    "self": "https://demo-api.directus.app/_/items/articles",
    "next": "https://demo-api.directus.app/_/items/articles?offset=40&limit=40",
    "last": "https://demo-api.directus.app/_/items/articles?offset=120&limit=40"
}
```

philleepflorence:
Okay cool, and for the number of rows returned when filtering - `filter_count`?

rijkvanzanten:
That works, but... it also kinda sounds like the amount of filters that were applied. Maybe `filtered_count`?

benhaynes:
Rijk and I just discussed this and he's updated his response above. The formatting/keys he has there are good to go!

philleepflorence:
So go with the following?
```
"limit": 40,
"page": 5,
"page_count": 20,
"filtered_count": 100,
"links": {
    "self": "https://demo-api.directus.app/_/items/articles",
    "next": "https://demo-api.directus.app/_/items/articles?offset=40&limit=40",
    "last": "https://demo-api.directus.app/_/items/articles?offset=120&limit=40"
}
```

rijkvanzanten:
Yes 💪

philleepflorence:
Cool!

rijkvanzanten:
Any updates on this? Eagerly awaiting this functionality in the API 🙂

philleepflorence:
I will release this Saturday or before.
Sorry for the delay.

philleepflorence:
@benhaynes and @rijkvanzanten : Since the `filtered_count` is now very important, should it be included in all filtered results response even if pagination is not?
Or should I just activate pagination in all responses since it is the same DB query that gets the `filtered_count` that also processes the `pagination`?
Updated the code to the new schema, will push once you confirm...

philleepflorence:
Current metadata response when the `limit` parameter is not provided, but a `page` parameter is provided:
```
{
    "meta": {
        "collection": "app_icons",
        "type": "collection",
        "result_count": 63,
        "total_count": 3040,
        "filter_count": 63,
        "limit": 1,
        "offset": 2,
        "page": 3,
        "page_count": 63,
        "links": {
            "self": "http://api.directus.local/_/items/app_icons",
            "first": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=1&offset=1",
            "last": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=63&offset=62",
            "next": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=4&offset=3",
            "previous": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=2&offset=1"
        }
    },
    "data": [...]
}
```

philleepflorence:
Metadata response with the `limit` and `page` parameters:
URL: `http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=2&limit=10`

```
{
    "meta": {
        "collection": "app_icons",
        "type": "collection",
        "result_count": 10,
        "total_count": 3040,
        "filter_count": 63,
        "limit": 10,
        "offset": 20,
        "page": 3,
        "page_count": 7,
        "links": {
            "self": "http://api.directus.local/_/items/app_icons",
            "current": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=2&limit=10",
            "first": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=1&limit=10&offset=10",
            "last": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=7&limit=10&offset=60",
            "next": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=4&limit=10&offset=30",
            "previous": "http://api.directus.local/_/items/app_icons?access_token=directus@admin&meta=*&filter[keywords][contains]=account&page=2&limit=10&offset=10"
        }
    },
    "data": [...]
}
```

bjgajjar:
@philleepflorence 

PR seems okay but there's one question which I need to ask you as well as @benhaynes  and @rijkvanzanten. Right now for the first page the `previous` link set with the `page` value "-1" and for the `last` page the `next` link is set with the `page` value "1". But I think we need to set the `page` as null here so if user want to make the `Next` and `Previous` button disable then they can do that easily.

Let me know your thoughts!

benhaynes:
In the example @rijkvanzanten made above, he excluded the first/previous when on page one. I'm fine with setting these to whatever, but the values should be consistent.

bjgajjar:
@benhaynes  Agreed! We have to exclude those values or need to set those as `null`. If @rijkvanzanten 
also agreed on that then @philleepflorence please apply the needful changes on PR :)

shealavington:
I would agree on null, otherwise, I believe you could end up with `Cannot read 'last' of undefined`. :roll_eyes: :joy:

rijkvanzanten:
> Since the filtered_count is now very important, should it be included in all filtered results response even if pagination is not?

`filtered_count` should be a separate `meta` value, that's separate from pagination.

philleepflorence:
I will update and push back online tomorrow...

philleepflorence:
Assumptions:

- `Limit` always takes precedence over the `page` parameter - so if the `result_count` is 10 and `limit` is 10, but `page` is 2, the `page` parameter would be corrected to 1.
- If there is only one page, (`page_count` = 1), `first`, `last`, `next`, and `prev` would be `null`.

rijkvanzanten:
Can we also make sure that

* `prev` is `null` when `page = 1`
* `next` is `null` when `page = pages.length`

philleepflorence:
Confirmed - will update later tonight if applicable.

philleepflorence:
CONFIRMED:

If you are on page 1 of 12, then `prev` should be `NULL` as you can't go to `0`
If you are on page 12 of 12, then `next` should be `NULL` as you can't go to `13`
If you are on page 1 of 1, then `next`, `prev`, `first` and `last` should be `NULL` as you can't go to `0` or `2`

PR #947, opened 2019-05-11
Fix 943

itsmerhp:
Two issues resolved : 

1. Bad SQL is generating while mysql reserved keywords used as field of collection and set it as primary key.
**Solution :** Escaped PK in where condition of query.
2. While converting mysql field to **CHAR**, length is set to 1
**Solution :** In required length field **CHAR** datatype is not set and due to that field length is passing 0 currently, so added **CHAR** in getTypesRequireLength function.


benhaynes:
Thanks @itsmerhp — @bjgajjar will review when she's back tomorrow. :)

Issue #945, opened 2019-05-10
v2.0.23 - Thumbnailer not working - giving 500 error


yarick2000:
# Bug Report

## Steps to Reproduce

1. Go to App and open File Library - all images without thumbnails
4. In browser developer panel you will see that all of them getting 500 error

## Expected Behavior
Thumbnails should be shown

## Actual Behavior
Thumbnails not shown, the default image shown instead

## Other Context & Screenshots

In API error logs there is a following message:
2019/05/10 14:04:41 [error] 1635#1635: *54 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Error: Call to undefined method Directus\Filesystem\Thumbnailer::getSupportedFileExtensions() in /mnt/c/Projects/Directus/directus-api/src/core/Directus/Filesystem/Thumbnailer.php:290
Stack trace:
#0 /mnt/c/Projects/Directus/directus-api/src/core/Directus/Filesystem/Thumbnailer.php(255): Directus\Filesystem\Thumbnailer->isSupportedFileExtension('jpg')
#1 /mnt/c/Projects/Directus/directus-api/src/core/Directus/Filesystem/Thumbnailer.php(60): Directus\Filesystem\Thumbnailer->extractThumbnailParams('casino/200/200/...')
#2 /mnt/c/Projects/Directus/directus-api/public/thumbnail/index.php(34): Directus\Filesystem\Thumbnailer->__construct(Object(Directus\Filesystem\Filesystem), Object(Directus\Filesystem\Filesystem), Array, 'casino/200/200/...')
#3 {main}
  thrown in /mnt/c/Projects/Directus/directus-api/src/core/Directus/Filesystem/Thumbnailer.php on line 290" while reading response header from upstream, client: 172.28.5.14, server: api.directus.com, request: "GET /thumbnail/casino/200/200/crop/good/casino-holdem.jpg HTTP/1.1", upstream: "fastcgi://unix:/run/php/php7.3-fpm.sock:", host: "api.directus.com:8080", referrer: "http://api.directus.com:8080/"

I was going to the code and replaced code of the isSupportedFileExtension function in Thumbnailer class:
    public function isSupportedFileExtension($ext)
    {
        return in_array(strtolower($ext), $this->getSupportedFileExtensions());
    }

 with the following one:

    public function isSupportedFileExtension($ext)
    {
        return in_array(strtolower($ext), Thumbnail::getFormatsSupported());
    }

[...]

PR #951, opened 2019-05-12
Fix 945

itsmerhp:
Current Issue of #945 is not reproducible, but I have found and resolved another issue in this task is, on uploading file an error is throwing(**Warning:**  strlen() expects parameter 1 to be string, array given in `/src/core/Directus/Database/TableGateway/BaseTableGateway.php` on line **406**)


rijkvanzanten:
This is a duplicate of https://github.com/directus/api/pull/933. Thanks though @itsmerhp!

PR #956, opened 2019-05-15
added collection/table to InvalidFieldException

Lapsus:
This PR adds the collection/table name to InvalidFieldException.

I find it very hard to find errors made in a "fields" definition, if only the field name is shown. Fields like "created_by" or "status" for example may be used multiple times. By adding the collection/table name to the exception message it is a lot easier to find errors in the definition.


benhaynes:
Thanks @Lapsus! I agree that the errors would be clearer with the collection name included. We'll review this ASAP.

PR #959, opened 2019-05-16
reuse item service instead of using a new instance

Lapsus:
I found that while fetching the logo and the fields in SettingsService a new instance of the ItemsService is used. But the ItemsService is also created during object construction and available.

So i refactored this to use the existing service. If there is no perticular reason to create a new instance i dont know of this should improve performance a litte.


benhaynes:
Thank you @Lapsus !! 😊

PR #969, opened 2019-05-21
Fix #968

### LINKED ISSUE #968: Creating item to "directus_user_roles" collection was denied when creating users
I'm trying to create a new user using a non-administrator role. I'm having and error `Creating item to "directus_user_roles" collection was denied`.
 if I go back to the user directory after the error, the users is created but without the role in place. Editing the same user result into the same error.

I already set the core permission to the role. Like so: 
image

I tried commenting the script below in **CoreServicesProvider.php** but it still didn't work.
 `if (!$acl->isAdmin()) {
                    throw new ForbiddenException('You are not allowed to create, update or delete roles');
                }`

My goal is allow non-admins to manage their club members on their own without the administrator assistance.
Is this possible on the application? Is there a misconfiguration or additional step to make this work?

itsmerhp:
> This Issue is taking place due to record insert / update in **directus_user_roles** collection restricted for administrator user only, even if other role user has permission for it.
> Removed additional restriction of administrator user only because role wise permission is already there.


benhaynes:
Thanks @itsmerhp! 

@rijkvanzanten always tells me to delete code instead of commenting it out... might be cleaner. :)

rijkvanzanten:
Yup, we have a full archive of every line of code that ever was and ever will be in Git. No need to keep a duplicate in the codebase itself too 😁

bjgajjar:
@itsmerhp 

Kindly delete the code instead of commenting it and push it again! I'll merge it. :)

itsmerhp:
> @itsmerhp
> 
> Kindly delete the code instead of commenting it and push it again! I'll merge it. :)

It's done.

PR #976, opened 2019-05-24
enables custom logo from directus_files when permissions on directus_files are set to mine or role

Lapsus:
pass the acl flag for the tableGateway factory from the settings endpoint to fetch a logo from directus_files even when the permissions are set to "mine" or "role".

The TableGateway already supports the circumvention of the acl. As the SessingsService uses the ItemService to fetch the logo i passthru an additional parameter to findByIds(). 

As you have to have the ID to fetch data i dont think this creates a security issue. Default the acl is in place, but if you have some ids (from settings for example) and want to fetch this items and ignore acl you now can do that.

The alternative in my opinion would be to have the custom logo as url instead of a directus_file.

### DIFF
--- a/src/core/Directus/Services/ItemsService.php
+++ b/src/core/Directus/Services/ItemsService.php
@@ -101,7 +101,9 @@ class ItemsService extends AbstractService
         $ids = StringUtils::safeCvs($ids, false, false);
 
         try {
-            $this->getAcl()->enforceRead($collection, $statusValue);
+            if (false !== $acl) {
+                $this->getAcl()->enforceRead($collection, $statusValue);
+            }
         } catch (ForbiddenCollectionReadException $e) {
             if (is_array($ids) && count($ids) > 1) {
                 throw $e;
--- a/src/core/Directus/Services/ItemsService.php
+++ b/src/core/Directus/Services/ItemsService.php
@@ -101,6 +101,8 @@ class ItemsService extends AbstractService
         $ids = StringUtils::safeCvs($ids, false, false);
 
         try {
+            // if acl check is disabled (e.g. fetching the logo from the settings endpoint/service) do not
+            // enforce permissions here!
             if (false !== $acl) {
                 $this->getAcl()->enforceRead($collection, $statusValue);
             }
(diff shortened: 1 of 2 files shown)


benhaynes:
Thanks @Lapsus!! I'll let @rijkvanzanten and @bjgajjar review and weigh in.

bjgajjar:
@Lapsus Without using this PR, I am able to get the `logo` in `settings` API after setting the `permission` to none for that role. So can you please elaborate what's the use case of this PR?

Lapsus:
@bjgajjar Did you set the permisisons on directus_files?

If you store the logo (in settings) as a directus_file, then this file (the logo) should always be fetched with the settings endpoint/service.

But in the current state if you for example set directus_files permisison to "none" or "mine" the logo will not be fetched for a non admin user. For a admin user it of course works, because all permissions are disabled for admins.

The PR utilizes the following:
The TableGateway factory already implements the disabling of acl checks. So when fetching the settings and the logo from the settings endpoint, a parameter is passed along to the tablegateway disabling acl checks for this query. And therefore you always can fetch the logo regardless of the permissions.

The alternative would be to put the logo outside of directus_files. For example by switching the logo value to a string value instead of a directus_file id.

bjgajjar:
@Lapsus  Yes, I set the permission on `directus_files` for the non-admin user!

I try to check with and without this PR. But the result is the same for me "Item Not Found".

So let me know what am I missing!

Lapsus:
Hmm ... you have a custom logo?
With the PR you should get the logo, without you get get Item Not Found.

bjgajjar:
We don't have `mine` permission in `directus_fields` table and for `none` permission this PR not works. So kindly fix it with `none` permission and send your PR again :)

src/core/Directus/Services/ItemsService.php
```diff
@@ -101,7 +101,9 @@ public function findByIds($collection, $ids, array $params = [], $acl = true)
         $ids = StringUtils::safeCvs($ids, false, false);
 
         try {
-            $this->getAcl()->enforceRead($collection, $statusValue);
+            if (false !== $acl) {
```
shealavington: > (false !== $acl)

This thoroughly annoys me to look at/read! :joy:

rijkvanzanten: wrong not is @shealavington (lets run back to front-end land shall we)

PR #978, opened 2019-05-27
prevented RelationalTableGateway to follow update on directus_user_roles relation

### LINKED ISSUE #977: Bug updating related tables using files and file interface
# Bug Report
When you use a files or file interface you also join users and roles to the collection.
During the update the RelationalTableGateway (line 622ff) also tries to update directus_user_roles and directus_roles which is only allowed for admin users (CoreServiceProvider line 643ff). 

## Steps to Reproduce
1. Create a collection with "user-created" interface and add a "files" interface. 
2. Autocreate relations. 
3. Add a role, grant "mine" permission to "select/update/delete". 
4. Upload a file, select the file in the "files" interface.
5. Click save.
 
## Expected Behavior
Item and relation should be saved.

## Actual Behavior
Error: you can not update item in directus_user_roles.

I would suggest excluding system tables like directus_user_roles, directus_user and directus_roles from the update process in the  RelationalTableGateway.

## Technical Details
manual install with latest api and app master branch.

Lapsus:
closes #977 

If you have a data collection that is m2m related to a directus_* table (for example directus_files), the RelationalTableGateway should stop searching for updated data at directus_user_roles.

Updating users or roles is only allowed for admin user and is handled by CoreServiceProvider. There for the RelationalTableGateway used by the ItemsService/Endpoint should not try to update this data as and stop updating related data here.

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -524,8 +524,8 @@ class RelationalTableGateway extends BaseTableGateway
 
         $statusField = $tableSchema->getStatusField();
         $logEntryAction = ArrayUtils::get($params, 'revert') === true
-                            ? DirectusActivityTableGateway::ACTION_REVERT
-                            : DirectusActivityTableGateway::ACTION_UPDATE;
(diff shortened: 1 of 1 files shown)


benhaynes:
Thanks @Lapsus — we'll take a look! 😄

PR #980, opened 2019-05-28
Fix 923

itsmerhp:
Till now, fields details are restricted to admin role only, I have removed that condition to allow it for all the roles.

But, due to removal of admin only condition, all the users with roles having permission for **Fields**(System Collection) will able to get fields details of all the collections even if users don't have read permission for those collections.

**Question :** Would you want me to add restriction that user can get fields details for a collection if user have read permission for that collection and read permission for **Fields**(System Collection)?
If no, then this PR is fine to merge.
If yes, then I do have to make changes in this PR.


bjgajjar:
@itsmerhp We should display the `field` as per the permission only. Otherwise, all the `fields` will be opened which is wrong. Let us know your thoughts here @benhaynes

benhaynes:
Hmm, so the options are:

1. Only Admin role can see field details (how it is now)
2. Base field detail access on permissions for `directus_fields` (this PR)
3. Base field detail access on permissions for `directus_fields` _and_ `[collection]:read`

Is that correct? Are these field details needed for requesting normal data (eg: field names in normal responses) or is it only for more specific info like `type`, `options` and `validation`?

How does a normal user see the `directus_fields.note` (eg: on the item detail page) right now if this info is admin only?

Which of those three do you recommend @bjgajjar?

bjgajjar:
@benhaynes  I don't think so the general user will require the `field` related data! But if still, we want to provide this functionality then the 3rd option looks okay to me.

Right now the nonadmin user is able to fetch the details related to the `fields` from the `collections` API> The `directus_fields.note` will get from `fields` object of `collections` API.

benhaynes:
Thanks Binal!

@rijkvanzanten — does option three sound good to you?

rijkvanzanten:
Yes, I vote 3. It doesn't make sense to be able read the fields for collections that you're not allowed to read themselves

benhaynes:
The team has voted! @itsmerhp — can we add the extra check to match option 3?

itsmerhp:
> The team has voted! @itsmerhp — can we add the extra check to match option 3?

Yes @benhaynes, I will apply necessary changes and update PR.

itsmerhp:
Hello @benhaynes, @bjgajjar and @rijkvanzanten 
I have applied the changes as per Option 3 for all the APIs of **fields** endpoint.

Further changes applied:
1) Read blacklist restriction has been applied while fetching fields list and individual field details.
2) In fields update and delete APIs, field write blacklist restriction has been applied.

Please let me know, if any further change needed.

benhaynes:
Thanks @itsmerhp!! @bjgajjar will review this asap! 😄

PR #1001, opened 2019-06-07
Add file 'format' query parameter to Thumbnail URLs

brocgailit:
Allows for thumbnailing of different formats from source image (like webp!) via:

`[thumbnail_url]?format=[format]`

Easy peasy lemon squeezy!

### DIFF
--- a/public/thumbnail/index.php
+++ b/public/thumbnail/index.php
@@ -26,12 +26,16 @@ try {
 $settings = \Directus\get_directus_thumbnail_settings();
 $timeToLive = \Directus\array_get($settings, 'thumbnail_cache_ttl', 86400);
 try {
+
+    parse_str($_SERVER['QUERY_STRING'], $queryParams);
+
     // if the thumb already exists, return it
     $thumbnailer = new Thumbnailer(
         $app->getContainer()->get('filesystem'),
         $app->getContainer()->get('filesystem_thumb'),
         $settings,
-        urldecode(\Directus\get_virtual_path())
+        urldecode(\Directus\get_virtual_path()),
+        $queryParams
     );
 
     $image = $thumbnailer->get();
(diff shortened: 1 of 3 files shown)


benhaynes:
Thanks @brocgailit!! We'll review this asap! 😄 

Hopefully it's not confusing that the `filename` will have an extension and so will the thumbnail (query string). Actually, now that I'm looking at our thumbnailer URL I noticed that it only supports unique filenames across the whole project... not good for when we support directories or multiple storage adapters. :/

@rijkvanzanten thoughts?

brocgailit:
An extension is added only if it differs from the format and is stored alongside the 'original' thumb. 

image

The previous PR using URL segments was a bit hacky (webp being served with jpg extensions, etc) and had the potential for unnecessary redundancy in file storage where `/jpeg/file.jpg` and `/file.jpg` would store two identical files.

I'm not up to date on the directory and multple adapters plans exactly, but it doesn't seem like this specific PR would affect that since it simply stores the formatted file as a sibling.

Open to suggestions on this!

rijkvanzanten:
I think it's less of an issue once we use file hashes for the name by default

bjgajjar:
Hey @brocgailit 

I have just one query regarding this. Can you please help me out? I am not able to get the thumbnail of uploaded `webp` image.

Is there anything which I am missing?

brocgailit:
@bjgajjar Can you describe more what you mean?

Below is a live example of usage:
image

You should be able to just add `?format=webp` to your thumbnail urls (files are stored using same methods).  The file gets stored as a sibling to the non-queried image and is served as `image/webp`.

bjgajjar:
Hey @brocgailit, 

I get the sibling but that image has the extension of both images. The parent one, as well as the, requested one. 

In your example; You might get the image name like Imperail_Organic_Yeast_Lab.jpg.webp
image

Can you please resolve this?

vickyshah129:
Do we have the same feature for general images? I want to retrieve webp format for original images and not just thumbnails. Any insight?

PR #1002, opened 2019-06-08
Fix 455

itsmerhp:
Implemented nested payload validation


benhaynes:
Thanks @itsmerhp! 😄

src/core/Directus/Database/Schema/Object/Field.php
```diff
@@ -491,6 +491,16 @@ public function isOneToMany()
...
     }
+    
+    /**
+     * Checks whether the relationship is MANY TO MANY
+     *
+     * @return bool
+     */
+    public function isManyToMany()
+    {
+        return $this->hasRelationship() && $this->getRelationship()->isOneToMany() && $this->getInterface() == "many-to-many" ? true : false;
```
bjgajjar: @benhaynes @rijkvanzanten  Is that okay if we implement the condition based on the interface? 

I think here we can implement the condition based on the `directus_relations` table instead of the interface. Cause for M2M; we have 2 entries of a junction table in `directus_relations` table and for the others, there is only single entry!

benhaynes: I leave this decision to Rijk,  but that makes sense to me.

rijkvanzanten: We can not filter based on the name of the interface. For example, `files` is also a m2m style interface, and there will most likely be more in the future

bjgajjar: Okay. As all of us are agreed that;

> We can not filter based on the name of the interface. 

So @itsmerhp kindly replace the condition of interface with relations table entry! Let me know if you need any help for the logic!

itsmerhp: @benhaynes @rijkvanzanten @bjgajjar 
I have applied required changes in PR. Please check it.

PR #1022, opened 2019-06-15
Docker configuration proposal

eldarc:
I'm proposing this Docker configuration since it's working smoothly for me.

What are some of the improvements:

- By bundling the Docker configuration with the main API repository we can leverage the `builder` container which is based on `composer`. After dependencies are installed they are just copied to the image. This setup can be moved to `directus-docker`, but in my opinion, new releases should be built with the `composer builder` directly from the source. Instead of downloading the release.

- Introduced dynamic environment injection to config files in the entrypoint script. This way Nginx and PHP configs can be adjusted through environment variables. For example, the Nginx listen port can be set.

- Modified the api configuration file to detect if the database is connected with a UNIX socket or HOST:PORT.

- PHP-FPM and NGINX are connected through a UNIX socket + files belong to the `www-data` user.

- Tweaked a bit the PHP-FPM and Nginx configuration.

- Configured supervisor for controlling both PHP-FPM and Nginx services.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.github
+.gitignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,83 @@
+#
+# Builder
+#
+FROM composer as builder
+
+# Copy source to builder
+WORKDIR /directus/
+COPY . /directus/
+
+# Install dependencies
+RUN composer install --ignore-platform-reqs
+
+#
+# Project
+#
+FROM alpine:3.9
+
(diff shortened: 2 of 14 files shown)


WoLfulus:
Hi @eldarc , sorry the delay.

We're about to merge a new way to load configs and it will make a lot more sense for working with docker. Also, to avoid reimplementing docker stuff inside each project we'll keep it under `directus/docker`.

Keep in mind that we're also working on bringing a development environment over here too (using directus/docker implementation instead).

@bjgajjar I think we can close this one

PR #1028, opened 2019-06-17
Fix 1027

itsmerhp:
Field length validation issue resolved for **File** interface


bjgajjar:
We should not implement the condition based on a specific type. 

https://github.com/directus/api/blob/792ca64533267e9418cc81cab7af1e34e652859f/src/core/Directus/Services/AbstractService.php#L507

So I added new pr #1029  which will remove the validation for relational fields as the relational fields will be validated at the store time at their collection.

PR #1036, opened 2019-06-20
Fix Warnings

Kinzi:
I got the following error attached to a JSON response from the API when I tried logging in with the app (resulting in a Server Error screen in the app) :

```
<b>Notice</b>:  Undefined index: fields in <b>/var/www/apidev/directus7/src/helpers/file.php</b> on line <b>115</b><br />
<br />
<b>Warning</b>:  Invalid argument supplied for foreach() in <b>/var/www/apidev/directus7/src/helpers/file.php</b> on line <b>347</b><br />
```

This patch wraps type checks around both issues

### DIFF
--- a/src/helpers/file.php
+++ b/src/helpers/file.php
@@ -112,11 +112,7 @@ if (!function_exists('append_storage_information')) {
             return $rows;
         }
 
-        if (empty($params['fields'])) {
-            $params['fields'] = '';
-        }
-
-        $fields =  get_files_fields($params['fields']) || '';
+        $fields = !empty($params['fields']) ? get_files_fields($params['fields']) : '';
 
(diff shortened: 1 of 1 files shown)


src/helpers/file.php
```diff
@@ -112,7 +112,11 @@ function append_storage_information(array $rows, array $params = [])
             return $rows;
         }
 
-        $fields =  get_files_fields($params['fields']);
+        if (empty($params['fields'])) {
+            $params['fields'] = '';
+        }
+
+        $fields =  get_files_fields($params['fields']) || '';
```
bjgajjar: I think we can use ternary operator here for better code quality.

` $fields =  !empty($params['fields']) ? get_files_fields($params['fields']) : '';`

Issue #1016, opened 2019-06-13
has filter seems broken


nstCactus:
# Bug Report

## Steps to Reproduce

Considering the following tables:

````
----------   --------------------   -------------
| movies |   | movies_directors |   | directors |
|--------|   |------------------|   |-----------|
|        |   | id               |   |           |
| id     | ← | movies_id        |   |           |
|        |   | directors_id     | → | id        |
----------   --------------------   -------------
````

On the movies table I have a M2M field names `directors` and related to the 
`directors` table.

I'd like to fetch all movies having more than one director. 
From the documentation :

> Using has will return items that has at least the minimum number given as 
> related items.

Hence I make the following request: `/items/movies?filter[directors][has]=1`.

## Expected Behavior

The request should return all movies having more than one director.  
The generated query should be something like:

````sql
SELECT movies.id, title
FROM movies 
    LEFT JOIN movies_directors ON movies.id = movies_directors.movies_id
GROUP BY movies.id
HAVING COUNT(movies_directors.id) > 1
````

## Actual Behavior

I got results but not what I expected. As I didn't understand how they were 
filtered, I checked the MySQL query log. Here is the generated query :

````sql
SELECT movies.id AS id, 
       movies.title AS title 
FROM movies 
WHERE movies.id IN (
    SELECT movies_directors.movies_id AS movies_id 
    FROM movies 
        RIGHT JOIN movies_directors ON movies.id = movies_directors.movies_id 
    WHERE movies_directors.id LIKE 1 
        OR movies_directors.movies_id LIKE 1 
        OR movies_directors.directors_id LIKE 1
) 
ORDER BY movies.id IS NULL, 
         movies.id ASC, 
         movies.id ASC 
LIMIT 200 OFFSET 0
````

### Simplified database dump (should be restorable on a running Directus instance)
````sql
INSERT INTO `directus_collections` (`collection`, `managed`, `hidden`, `single`, `icon`, `note`, `translation`)
VALUES ('directors', 1, 0, 0, NULL, NULL, NULL),
       ('movies', 1, 0, 0, NULL, NULL, NULL),
       ('movies_directors', 1, 1, 0, NULL, 'Junction Collection', NULL);
INSERT INTO `directus_fields` (`collection`, `field`, `type`, `interface`, `options`, `locked`, `validation`, `required`, `readonly`, `hidden_detail`, `hidden_browse`, `sort`, `width`, `group`, `note`, `translation`)
[...]

PR #1041, opened 2019-06-20
[WIP] Fix has filter

nstCactus:
My attempt at fixing #1016.

This is what I got so far. 

Given the request `items/movies?fields=*&filter[directors][has]=2`, it generates the following query :

````sql
SELECT movies.id AS id, movies.title AS title
FROM movies
WHERE movies.id IN (
    SELECT movies_directors.movies_id AS movies_id
    FROM movies
        RIGHT JOIN movies_directors ON movies.id = movies_directors.movies_id
    GROUP BY movies.id
    HAVING COUNT(movies.id) >= '2')
ORDER BY movies.id IS NULL, movies.id ASC, movies.id ASC
LIMIT 200 OFFSET 0
````

But it breaks related fields selectors, and most probably a lot more.

@bjgajjar Do you have any advice on how to improve this?

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -1546,14 +1546,13 @@ class RelationalTableGateway extends BaseTableGateway
             $tableSchema = SchemaService::getCollection($relatedTable);
             $relatedTableColumns = $tableSchema->getFields();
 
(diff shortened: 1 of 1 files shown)


bjgajjar:
@nstCactus 

I tested this. This code will only affect O2M fields, And the filters of O2M works fine with this PR.

The proper mentions of the broken scenarios will be really appreciated. Otherwise, this PR seems okay to me :)

PR #1048, opened 2019-06-23
Adds support for different config sources

WoLfulus:
This PR adds support for different configuration sources like environment variables, json files and php files. It also introduces configuration schemas to automatically cast types from sources that only supports strings (like environment variables)


WoLfulus:
I made a new PR since the other one was messed up with rebase stuff, this is much cleaner after I squashed the commits

I think we can make a review on this now. 

I've added an example (it won't break how `Config` actually works, the idea is that we're just supporting different config sources)

Ther are some tests here [https://github.com/directus/api/blob/897e4ce60cc2833e6d49aef3c09e2dda001de5c9/tests/api/Config/ConfigTest.php](https://github.com/directus/api/blob/897e4ce60cc2833e6d49aef3c09e2dda001de5c9/tests/api/Config/ConfigTest.php)

Keep in mind that the `Config` class is exactly the same, we'll just change how we feed data into it.

Once the reviewers agree on the implementation, we can proceed integrating the changes into `core` to conditionally load context either from `api.php` or `environment variables`

Also remember that as we discussed earlier, we're assuming docker environment will only support one `tenant` / `project` per container. So in order to have multiple projects, different instances should run behind a reverse proxy with different environment variables (if the operator wants only one container running, the environment variable loading should be disabled and volume mapped files should be used instead).

benhaynes:
Thank you @WoLfulus!!! We'll review this ASAP! 😄 

@bjgajjar @rijkvanzanten — let's take a look and make sure this is good for both of you.

WoLfulus:
It's not ready for merging just yet: we need a review before I proceed integrating it into core and decide how we`ll dynamically select which config source to use.

We can discuss this in here: my first thought was to just check for `DIRECTUS_USE_ENV` variable and make it the default behavior if it's present.

src/core/Directus/Application/Application.php
```diff
@@ -142,7 +142,7 @@ public function setCheckRequirementsFunction(\Closure $function)
     protected function createConfig(array $appConfig)
     {
         return [
-            'settings' => ArrayUtils::pull($appConfig, 'settings', []),
+            'settings' => $appConfig['settings'],
```
bjgajjar: May I know why you remove the condition of considering the default value? Currently, It is throwing the error of undefined settings in the API. Kindly revert this change if this is not necessary or add `isset` condition.

WoLfulus: That call is removing settings from config, which was being used in logger (it never worked to be honest). I'll check another approach.

Also, what files do you mean? Would you mind listing them? Should I just *remove* them?

PR #1084, opened 2019-07-04
Issue Fix #991

### LINKED ISSUE #991: [SECURITY] Insufficient Anti-Automation – Brute Force Attack
# Feature Request

About audited Directus version. 
It has been cloned from suite repo.
Latest commit https://github.com/directus/directus/commit/1d151a9034514e3f2ec1c80001e7c5fffdef2d4e

## Description:

Insufficient Anti-automation occurs when a web application permits an attacker to automate a process that was originally designed to be performed only in a manual fashion, i.e. by a human web user.

## Business risk:

Once exploited, attackers can use an automatic tool or a simple script that performs a brute-force authentication attempts, and possible cause a legitimate user lock out.

## Technical details:

The login mechanism of the application does not prevent a brute-force on login attempts and on change password function.

Fixes security hole.

* Implement an anti-automation mechanism such as CAPTCHA
Note: do not use account locking, because of the ability to perform a DoS on user accounts.
* Complementary instructions for protecting brute-forcing on users can be found in the following URL:
http://projects.webappsec.org/w/page/13246938/Insufficient%20Anti-automation
https://www.owasp.org/index.php/Blocking_Brute_Force_Attacks

Maybe, with help/guidance from Directus team.

bjgajjar:
User can set maximum `login_attempts_allowed` from global settings. If the user reached this limit then their account will be `suspended` and then need to contact admin to reactivate the account. Once admin reactivates the accounts or the user logged in successfully in the system this limit will be reset for the user.

@rijkvanzanten I am throwing the error message from `API` after max login attempt reached that "Your account is suspended due to the maximum allowed login attempts. Please contact your administrator to activate your account." but `APP` is displaying `User Inactive`. Please check that at your end once. 

We can add below points as a feature request.

- User will be `blocked` for `n` number of minutes and after that they able to log in again without contacting admin.

-  User will able to set  `login_attempts_allowed` for `n` number of minutes after this time limit reached they will get the same `login_attempts_allowed` again.


benhaynes:
Thanks @bjgajjar !! Has this been added to the docs as well? We need to make sure that all new/updated features are added to the docs as we go.

PR #1086, opened 2019-07-04
Issue Fix #1075

### LINKED ISSUE #1075: After passing filter_count, Do not query to return the fields limit links offset page_count and page
# Feature Request
Do not query to return the fields `limit ` `links` `offset` `page_count` `page`
There may be other reasons, but for me, I don't need them.
image

These fields are of no value to me, and I can get them all through JS calculations.
` limit ` , ` offset `: I define them myself, and I confirm that they are known and do not need to return.
`links` `offset` `page_count` `page`:  I can get them through JS splicing calculation and other methods

After passing filter_count, Do not query to return the fields `limit ` `links` `offset` `page_count` and `page`

urvashithakar:
Remove page related data when passing `filter_count` in meta param.


benhaynes:
@bjgajjar — let's make sure this response format and metadata options are all documented. 😊

PR #1091, opened 2019-07-05
Issue Fix #1072

### LINKED ISSUE #1072: Updates in relational interfaces require required fields
When updating an existing item through a relationship, like a one to many, the API requires all required fields to be set, even if they already have a value in the database.

Lets say I have a collection named `sections` that has a one to many to `cards`. When I update existing card with `id = 4` by sending the following payload

```json
{
  "cards": [
    {
      "id": 4,
      "note": "Test"
    }
  ]
}
```

the API will return

```json
{
  "error": {
    "code": 4,
    "message": "status: This value should not be blank."
  }
}
```

even though the field already has a status value in the DB, and it's not a newly created item.

When updating the item through the `/items` endpoint directly, this problem doesn't occur.

👉 The API should treat updates to existing items in relational fields the same as regular updates.

bjgajjar:
For M2M related collection's records were not updating. This issue is also resolved with this PR :)

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -608,7 +608,7 @@ class RelationalTableGateway extends BaseTableGateway
             }
 
             // Update/Add foreign record
-            if ($this->recordDataContainsNonPrimaryKeyData($foreignRow, $foreignTableSchema->getPrimaryKeyName()) && ($field->getType() != "file" && !isset($parentRow[$fieldName][$primaryKey]))) {
+            if ($this->recordDataContainsNonPrimaryKeyData($foreignRow, $foreignTableSchema->getPrimaryKeyName()) && $field->getType() != "file" ) {
                  // NOTE: using manageRecordUpdate instead of addOrUpdateRecordByArray to update related data
                 $foreignRow = $ForeignTable->manageRecordUpdate($foreignTableName, $foreignRow);
             }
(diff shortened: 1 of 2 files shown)


src/core/Directus/Database/TableGateway/RelationalTableGateway.php
```diff
@@ -608,7 +608,7 @@ public function addOrUpdateManyToOneRelationships($schema, $parentRow, &$childLo
             }
 
             // Update/Add foreign record
-            if ($this->recordDataContainsNonPrimaryKeyData($foreignRow, $foreignTableSchema->getPrimaryKeyName()) && ($field->getType() != "file" && !isset($parentRow[$fieldName][$primaryKey]))) {
+            if ($this->recordDataContainsNonPrimaryKeyData($foreignRow, $foreignTableSchema->getPrimaryKeyName()) && $field->getType() != "file" ) {
```
hemratna: Find a way to remove hardcoded condition like `$field->getType() != "file"`
Adding comment in TODO.

src/core/Directus/Services/ItemsService.php
```diff
@@ -171,7 +171,113 @@ public function findOne($collection, array $params = [])
...
+                        }
+                    }
+                    $this->validatePayload($relationalCollectionName, null, $validatePayload,$params);
+                }
+            }
+        }  
+    }
+    
+    /**
+     * Validate Alias Collection Fields
```
hemratna: As this function adds support for OneToMany, ManyToOne. Please mention in comment.

PR #1093, opened 2019-07-06
#1072

itsmerhp:
Applied change in the nested payload validation of Update API


itsmerhp:
Issue: In M2M related collection's records were not updating.
Solution: Change applied in `addOrUpdateManyToOneRelationships` to resolve it.

hemratna:
Thanks, @itsmerhp for PR.

@bjgajjar sent a PR #1091 with batter approach (which handles M2M too).

PR #1094, opened 2019-07-08
Passing filter_count when no filter is passed.

urvashithakar:
If endpoint have a `filter_count` as meta param and dont have a `filters` then the API was not returning the `filter_count` in the response as it is based on the filters only. And if you are not passing filters then there is no need to return the counter for it. But  that may create the issue at APP side. They need to add the condition for displaying the counters.

So with this PR; user will able to get `total_count` as `filter_count` if they has been passed `filter_count` as meta and not passing any filter. Otherwise the filter_count will work as expected(Based on filters) 🙂


cdwmhcc:
It's word! Thank you very much!

PR #1096, opened 2019-07-08
Fixes log and config access keys

WoLfulus:
Fixes problems raised by @hemratna and @bjgajjar

### DIFF
--- a/config/api_sample.php
+++ b/config/api_sample.php
@@ -8,7 +8,7 @@ return [
 
     'settings' => [
         'logger' => [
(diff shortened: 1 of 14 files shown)


hemratna:
@WoLfulus For the S3 the `options` variable is missing reflected. Due to this all the images uploaded are not accessible from public URL.

WoLfulus:
This might be a config related issue since I can't reproduce it.
Please send me over slack both your `config` folders so I can take a look what is causing it.

Thanks!

PR #1112, opened 2019-07-15
Add lang parameters in GraphQL user collections.

### LINKED ISSUE #1110: Add lang filter to GraphQL collection when translation is inside the content.
# Feature Request
Hi to all, I'm seeing that there isn't a way for query content based on language in Directus via GraphQL, I think the better way for doing this is to add `lang` filter when the content has some `translation` filed in and transform this filter work like `lang` filter like REST API.

Pay attention, if the field has multiple `translation` fields we cannot select a different language on the same content, but I think this behavior is what we want when we consume translated content, simply define the language a get it.

Allow better GraphQL query for translated content, using always the same filter as happen for REST API.

Add `lang` string filter to content when they have `translation` filed in his definition.

I will try to work on this.

Bedotech:
Fix #1110 

Now GraphQL query can use lang parameter for filter user collections by
languages, the changes are reflected in the schema.

### DIFF
--- a/src/core/Directus/GraphQL/Collection/CollectionList.php
+++ b/src/core/Directus/GraphQL/Collection/CollectionList.php
@@ -12,6 +12,7 @@ class CollectionList
     protected $limit;
     protected $offset;
     protected $container;
+    protected $lang;
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


benhaynes:
Thanks @Bedotech! We'll review ASAP! 😄

PR #1113, opened 2019-07-15
Add cache for GraphQL content

Bedotech:
Directus only cache content that coming from a GET request, this behavior
works well for REST but GraphQL put the payload of a query in the body
encoded as JSON.

This is a temporary workaround, in fact, the cache doesn't consider ACL
and mutation but this really make usable GraphQL integration, because
without this is really slow (above 1.5 seconds) for a simple query.

Consider this an MVP through full GraphQL implementation in Directus.

It's just a suggestion for usability I don't bother if you discard this!


benhaynes:
Thanks @Bedotech — we'll take a look!

hemratna:
Hi @Bedotech, Thanks for PR 🙌 
This is good start for adding support for caching in GraphQL.

PR #1115, opened 2019-07-15
GraphQL: add equality checks for strings

ale-rinaldi:
This adds the definition for the `eq` and `neq` filters for strings.


benhaynes:
Thanks @ale-rinaldi!

PR #1116, opened 2019-07-16
Disable error reporting until known configuration

WoLfulus:
This disableds error reporting until we know which environment we're in, and sets default value for user settings


WoLfulus:
@bjgajjar this should do it.

PR #1119, opened 2019-07-17
Fixed possible null pointer exception in fetchItems when no ACL is used

christianrr:
Fixed an issue, that fetchItems raises a null pointer exception on when a table gateway with no ACL is used (set to 'false'). `$this->acl` is null in this case, subsequently `$this->acl->getCollectionStatuses($this->table)` throws an exception.


benhaynes:
Thanks @christianrr — we'll review ASAP... 👍

PR #1123, opened 2019-07-18
create thumb for pdf if imagick is available

iiome:
Solving #1081 

Reusing the code in Filesystem/ that existed (in previous version?)

Only works when the imagick module is installed.
When imagick is not installed the current behaviour is preserved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@directus/api",
   "private": true,
-  "version": "2.2.2",
+  "version": "2.3.0",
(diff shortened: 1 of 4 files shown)


benhaynes:
Thank you @iiome! Will review this ASAP...

hemratna:
The thumbnail is generating but `files` endpoint return blank for `thumbnails` key in case of `PDF`. 
We need to add the `thumbnails` for PDF too.

PR #1124, opened 2019-07-18
Added file support for 7.0 (Explained)

shealavington:
Having `public` in front of `const` completely breaks the application for `PHP 7.0` usage, which broke everything when I pulled origin.

Though I understand `PHP 7.0` isn't officially supported, and that `PHP 7.1+` is, there is no reason to use public alongside const as the default visibility of class constants are public. We might as well provide support where possible if it doesn't hurt.

Also explained here:
https://stackoverflow.com/a/51568547

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@directus/api",
   "private": true,
-  "version": "2.2.2",
+  "version": "2.3.0",
   "description": "Directus API",
   "main": "index.js",
   "repository": "directus/api",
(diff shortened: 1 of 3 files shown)


benhaynes:
Thanks Shea! If this doesn't break anything or reduce performance for supported versions then I agree that it's nice to get things working with v7+.

Thoughts @directus/api-team ?

shealavington:
Indeed, from what I've read, `const`'s are public by default anyway, so declaring them as `public` wouldn't make a difference to their default behavour. In 7.1 the ability to make them `protected and private` was added.

Just wanted this checked :stuck_out_tongue_closed_eyes:

Issue #1121, opened 2019-07-17
Many to one  Cannot add new items


cdwmhcc:
# Bug Report
Many to one  Cannot add new items

## Steps to Reproduce

1. Create a collection blog, add field title, content
2. Create a collection blog_like and add a field `blogid` (many to one)
3. Setting Public Permissions
4. Call the interface to add like
 ```
this.$client.createItem("blog_like",{blogid: 1}).then((res) => {
   console.log(res)
 })
```
4.See error
> Statement could not be executed (23000 - 1048 - Column 'action_by' cannot be null)

![image](https://user-images.githubusercontent.com/8030134/61401022-20fe2280-a904-11e9-908c-4f229a20b633.png)

## Expected Behavior
Many to one  Cannot add new entries

## Actual Behavior
It used to work, but I don't know why it suddenly stopped working.
Maybe because of the update of jsSDK, maybe the problem of api, maybe the problem of app. 
Hope to be fix

## Technical Details

- Device: [Desktop]
- OS: [Windows 10.1809]
- Web Server: [Nginx:1.17.0]
- PHP Version: [7.3.6]
- Database: [5.7.26]
- API Version: [2.3.0]
- APP Version: [7.7]
- Install Method: [cloned `master` branch]

bjgajjar:
@cdwmhcc 

May I have a dump SQL file?

bjgajjar:
Fixed in #1126

cdwmhcc:
> Fixed in #1126

Thank you!

Issue #1098, opened 2019-07-09
Ignore token for public endpoints


rijkvanzanten:
When you request a public endpoint with an invalid token, the API will return an error. Seeing that this data is public to begin with, it shouldn't cause problems. The API should ignore the token for public endpoints.

Ref https://github.com/directus/app/issues/1821

bjgajjar:
Fixed in https://github.com/directus/api/pull/1111

PR #1126, opened 2019-07-18
#1121

itsmerhp:
**Issue:** while creating an item for Public role, an exception thrown
**Solution:** For Public role, UserId should be set to `0` instead of `NULL`

### DIFF
--- a/src/core/Directus/Application/Http/Middleware/AuthenticationMiddleware.php
+++ b/src/core/Directus/Application/Http/Middleware/AuthenticationMiddleware.php
@@ -53,7 +53,8 @@ class AuthenticationMiddleware extends AbstractMiddleware
 
             if (!is_null($user)) {
                 $rolesIpWhitelist = $this->getUserRolesIPWhitelist($user->getId());
-                $permissionsByCollection = $permissionsTable->getUserPermissions($user->getId());
+                $permissionsByCollection = $permissionsTable->getUserPermissions($user->getId());                
+                $hookEmitter->run('auth.success', [$user]);
             } else {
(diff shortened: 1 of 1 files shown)


src/core/Directus/Application/Http/Middleware/AuthenticationMiddleware.php
```diff
@@ -102,9 +93,15 @@ public function __invoke(Request $request, Response $response, callable $next)
...
-
+       
+        // TODO: Adding an user should auto set its ID and GROUP
+        // TODO: User data should be casted to its data type
+        // TODO: Make sure that the group is not empty
+        $acl->setUserId($user->getId());
+        $acl->setUserEmail($user->getEmail());
+        $acl->setUserFullName($user->get('first_name') . ' ' . $user->get('last_name'));
+
+        $hookEmitter->run('auth.success', [$user]);
```
bjgajjar: @itsmerhp 

This code will break when the user has an invalid token. Kindly check that scenario and update the code accordingly. :)

hemratna: @itsmerhp Please refer the issue #1098

itsmerhp: @bjgajjar @hemratna 
I have committed the change and tested it for the below cases : 

1. Public endpoint
-  Without token
- Valid token of another role user
- Invalid token

2. Non-Public endpoint
- Without token
- Valid token of another role user
- Invalid token

PR #1146, opened 2019-07-23
Issue Fix #1109

### LINKED ISSUE #1109: [SECURITY] Weak Password Policy
# Feature Request

About audited Directus version. 
It has been cloned from suite repo.
Latest commit https://github.com/directus/directus/commit/1d151a9034514e3f2ec1c80001e7c5fffdef2d4e

## Description:

The application does not enforce an efficient password policy on the password and security questions. As a result, it increases the chances that users select simple passwords, which could be easily enumerated in credential enumeration attacks (such as brute force and dictionary attacks).

## Business risk:

An attacker can easily guess weak password with brute force attack.

## Technical details:

The application does not enforce an efficient password policy on its users and enables them to select simple password combinations which could be easily enumerated.

The flaws in the policy enforcement originate from the following reasons:

* Users are not required to select passwords with enough character variations.
* Users are not required to select passwords that differentiate from the rest of their credentials.
* The password policy is not enforced in the reset password mechanism. During the test, we were able to change a user password to `a`:

screenshot

screenshot2

Fixes security hole.

The application should enforce an efficient password policy, which is suitable to the application sensitivity level. The password policy should be enforced according to the following guidelines:
* Set a minimum length of 8 characters for passwords.
* Each password must contain three different types of characters. It is recommended to limit the selection to uppercase and lowercase letters, and digits.
* The password must be different from the username, user ID or any other user credential.
* Enforce multi-factor authentication registration.
* The password policy must be enforced in any phase that enables password selection or alteration, including the user creation phase, the user registration phase, the password initialization phase and the password alteration phase.
* Enforce users to change their password once every quarter.

Maybe, with help/guidance from Directus team.

bjgajjar:
This PR will contain 4 options for Password Policy

1. None: No Regex
2. Weak: Regex for min length validation of 8 characters
3. Strong: Regex for
      - Minimum length of 8 characters
     - At least 1 small-case letter
     - At least 1 capital letter
     - At least 1 digit
     - At least 1 special character

4. Other: which have allow_other value right now.

Hey @rijkvanzanten,

I need your help with the Other option. As @benhaynes suggested here; I added one option named Other which contains allow_other value but may I know that the interface which Ben mentioned is already generated or we need to develop it? Let me tell you here; right now on the selection of Other I am not getting any text box although the value is allow_other.

Can you please guide me what should be there from API for development of Other? Right now the first 3 points are developed just the Other option is left.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,10 @@ composer.phar
 composer.lock
 /vendor
 
+# Exclude env vars and custom deployment scripts
+.env
+deploy.*
+
 # Ignore configuration files
 /config/*
 !/config/migrations.php
--- a/migrations/upgrades/schemas/20190722110232_password_validation_setting_field.php
+++ b/migrations/upgrades/schemas/20190722110232_password_validation_setting_field.php
@@ -14,7 +14,7 @@ class PasswordValidationSettingField extends AbstractMigration
             'type' => 'string',
             'note' => 'Weak : Minimum length 8; Strong :  1 small-case letter, 1 capital letter, 1 digit, 1 special character and the length should be minimum 8',
             'interface' => 'dropdown',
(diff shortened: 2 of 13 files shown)


rijkvanzanten:
please open a feature request for that on the app repo 👍

bjgajjar:
@rijkvanzanten 

Added one feature request for `Other` at APP side.  Currently, I removed the option of `Other` from **Password Policy**. Once you add it in `APP` repo; let me know so I can add it to here too.

PR #1150, opened 2019-07-24
Fix 1121

itsmerhp:
**Issue:** while creating an item for Public role, an exception thrown
**Solution:** For Public role, UserId should be set to `0` instead of `NULL`


bjgajjar:
@itsmerhp 

It seems like we have this code in develop branch already. Please check [here](https://github.com/directus/api/blob/develop/src/core/Directus/Application/Http/Middleware/AuthenticationMiddleware.php#L56).

We had been faced with some merge issue. And that's why this happens. Just check once with rebasing develop branch. It is already resolved.

Thanks for PR :)

PR #1151, opened 2019-07-24
Parent + Nested validation changes

itsmerhp:
**Issue 1(Drag column in listing page of APP):** **findByIds** is returning value as **Object** for JSON datatype fields, so that it is creating an issue in parent validation and nested payload validation.
**Solution:** Converted Object into the array for JSON fields.

**Issue 2:** Let's say a collection has JSON field and it is required, in update item remove value from it, for blank JSON field NULL value is passing, so that `isset` is not working for NULL value and creating issue in validation.
**Solution:** Used `array_key_exists` in place of `isset`

### DIFF
--- a/src/core/Directus/Services/ItemsService.php
+++ b/src/core/Directus/Services/ItemsService.php
@@ -12,6 +12,7 @@ use Directus\Util\StringUtils;
 use Directus\Validator\Exception\InvalidRequestException;
 use Zend\Db\TableGateway\TableGateway;
 use Directus\Database\SchemaService;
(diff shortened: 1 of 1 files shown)


bjgajjar:
Hey @itsmerhp 

It seems like we have this code in develop branch already. Please check [here](https://github.com/directus/api/blob/develop/src/core/Directus/Services/ItemsService.php).

We had been faced with some merge issue. And that's why this happens. Just check once with rebasing develop branch. It is already resolved.

Thanks for PR :)

PR #1163, opened 2019-07-30
added array type to schema

### LINKED ISSUE #1118: item.create and item.update hooks not executed
# Bug Report

## Steps to Reproduce

1. `git clone https://github.com/directus/api .`
2. `composer install`
3. Copy the old version of the configuration file(api.php) to config/
4. Create test item create hook, either as `item.create` or `item.create.TABLE_NAME`
 ````
  'actions' => [
    'item.create' => function (array $data) {
      $client = new \GuzzleHttp\Client([
        'base_uri' => 'https://example.com'
      ]);

      $data = [
        'type' => 'post',
        'data' => $data,
      ];

      $response = $client->request('POST', 'alert', [
        'json' => $data
      ]);
    }
  ]
````

## Expected Behavior
The endpoint should be called

## Actual Behavior
The code is never executed and will therefore not call the endpoint.

## Technical Details

- Device: Desktop
- OS: MacOS 10.14.5
- Web Server: Apache 2.4.37
- PHP Version: 7.2.0
- Database: MySQL 8.0.12
- Install Method:  cloned `master` branch

### PR DESCRIPTION
**Pull request by Lapsus**
fixed #1118 

Also reduced some duplicate code and added a constant for type Array.

### CODE REVIEW

**Comment by bjgajjar** [comment]
@WoLfulus 

It seems okay for `API` purpose. Kindly check for the `Docker` and let me know so it can be getting merged.

**Comment by Lapsus** [comment]
@WoLfulus I just found another issue with the schema validation.

I use the config in api.php to extend the service configuration. 
In my hooks i use for example a redis connection to write data. This connection uses the regular slim service configuration. Without the schema validation the service configuration is passed along. now it gets filtered out. And as the services configured there are not related to directus it makes no sense to add it to the schema config. Maybe we could validate the config if it is taken from env but skip this when the config is taken from the api.php?

File: src/helpers/app.php [hunk 1]
```diff
@@ -100,9 +100,7 @@ function get_project_config($name = null, $basePath = null)
             if (!file_exists($configFilePath)) {
                 throw new UnknownProjectException($name);
             }
-            $configData = $schema->value([
-                "directus" => Context::from_file($configFilePath)
-            ]);
+            $configData = Context::from_file($configFilePath);
```
WoLfulus: Completely skipping the schema for files would likely cause a lot of problems (and docker issues) if someone implement a feature on the api without knowing that the schema exists and it fails the code review. It won't load from environment variables even though the api.php file does actually load.

src/core/Directus/Config/Schema/Value.php
```diff
@@ -50,19 +48,20 @@ public function value($context)
...
-        default:
-            return $value;
+            case Types::INTEGER:
+                return intval($value);
+            case Types::BOOLEAN:
+                $value = strtolower($value);
+                return $value === "true" || $value ===  "1" || $value === "on" || $value === "yes" || boolval($value);
+            case Types::FLOAT:
+                return floatval($value);
+            case Types::ARRAY:
```
WoLfulus: Would you mind adding some tests for this on `tests/api/Config` dir? :)

WoLfulus: Would you mind adding some tests for this on `tests/api/Config` dir? :)

src/helpers/app.php
Lapsus: i disagree, for a few reasons.
1. this is how it was before. using the config from api directly.
2. configuring individual services, or a closure function as a hook can not be done via environment. as it is code.
3. you can not provide and foresee what options can be put in the config. and if you soften your schema validation with an ANY type, you could probably just skip it a all.

But this is just my opinion on the matter of course.

Lapsus: If i understand correctly, you are trying to enable configuration via environment, correct?
This is makes it easy to configure database connection etc. and pass it to the docker container in the yml, right?

But there are a lot more options in the api.php that you can configure. And a lot of this things can not be moved to a yml file, as they are actual code. For example a hook or a service factory. So i am not shure if it is the right approach to put the complete config from api.php to a schema and have it handle both ways.

Or am i completely wrong here?

WoLfulus: > this is how it was before. using the config from api directly.

Yes, but we must cover the current project needs too (which is loading settings from different source other than php files).

> configuring individual services, or a closure function as a hook can not be done via environment. as it is code.

This is one of my concerns on how configs are done right now. I'd completely separate hook setup from configs itself as configs we expect only values, and having an external hooks setup would allow us to make proper automated installations like `directus hook install` commands on custom docker images.

> you can not provide and foresee what options can be put in the config. and if you soften your schema validation with an ANY type, you could probably just skip it a all.

This would just allow custom configs to be loaded even though they are not in the schema itself. And you'd be responsible for the proper conversions inside the hooks.

> Keeping the schema for transforming values from .env files is fine of course. And actual code like factory functions for services or hooks will probably never work in a .env file anyway. So it does not matter there.

I agree with that, but as I mentioned before, this is a scenario that I'd refactor to something else and these should be done in a major release because it will break backwards compatibility.

WoLfulus: I think I messed up where I did the reply D:

src/core/Directus/Config/Schema/Value.php
Lapsus: will do tomorrow.

src/helpers/app.php
Lapsus: hmm and how about a hierarchical approach? 
Fetch the config from api.php and fetch .env when they exist and use it just for the relevant parts of the config. the .env options then replace the ones from the api. 

This way the schema also just has to over the .env part.?
Should work for both scenarios in my opinion.

WoLfulus: The problem is that we won't have a .env file most of the time (it's not how docker works). It should be fetched literally from worker process environment variables. That's why using `phpdotenv` you mentioned before won't work.

Hierarchical you mean .php taking priority and we override the values based on what's present on the environment variables?

That's another approach, but will be as expensive as the current setup because it means we still need to know what to search for on environment variables (also we won't know the data type of the variable to do proper conversions since all of them comes as strings - which is why we introduced the config "schema")

I thought about all of this earlier on the implementation, and the closest thing I came up as a solution was some kind of config factory (which is the `Context` class and `from_` methods). I haven't validated or put much thought into this but I tried making schema walk O(n) with `n` being the number of variables set in the schema.

benhaynes:
Hey @WoLfulus and @Lapsus — I know ya'll were going to chat on the Directus Slack regarding this... any updates? :)

Lapsus:
We are still discussing ... sorry Ben. I know it is an urgent matter. If you merge the PR the bugs are fixed, but it is not a permanent solution.

benhaynes:
No worries — just trying to stay on top of things. I appreciate you both taking the time to find the ideal solution!

Lapsus:
@WoLfulus 
for a compromise i added a new optional key "ext" to the schema and api.php. Type is array so one can put what ever config he wants in addition to the directus config under this key. This makes the schema pass as defined and is extensible for every need. 

I also added some tests as requested.

PR #1168, opened 2019-07-31
Add option to prevent upsizing contained thumbnails

wcerfgba:
I am using the thumbnailer to load all images to provide sanitization of user uploaded content -- I would rather pass the image through Intervention than serve the original upload to visitors, in case it was malicious.

Currently the thumbnailer forces the requested size, and will scale up images, but for this use case I would rather not scale up images. The resize method we use has a `upsize` constraint which prevents scaling the image [1].

I've exposed this as an option so it's backwards compatible.

I had a look for tests but couldn't find any. Do we want tests on the thumbnailer endpoints, or unit tests?

Thanks!  😀

[1] http://image.intervention.io/api/resize


wcerfgba:
Sorry @bjgajjar I originally branched from master by mistake, this is now fixed. :+1:

PR #1181, opened 2019-08-07
Issue fix #1179

### LINKED ISSUE #1179: Make settings->logger->path configurable through /projects
The `/projects` endpoint allows you to configure a lot of the fields in the config file, except for the `settings->logger->path` variable. This combined with https://github.com/directus/api/issues/1178 results in the fact that newly created projects may or may not be able to write logs.

https://github.com/directus/api/blob/f542f2f458825709dcb9e9ebe3ceb39e5af5350d/src/core/Directus/Util/Installation/InstallerUtils.php#L873-L939

I've tagged it both a bug and a feature request, as the `/projects` endpoint is pretty useless if a newly created project can't work.

bjgajjar:
Expecting `logs_path` in `API`

### DIFF
--- a/src/core/Directus/Util/Installation/InstallerUtils.php
+++ b/src/core/Directus/Util/Installation/InstallerUtils.php
@@ -894,6 +894,7 @@ class InstallerUtils
             'feedback_token' => sha1(gmdate('U') . StringUtils::randomString(32, false)),
             'feedback_login' => true,
             'timezone' => get_default_timezone(),
+            'logs_path' => __DIR__ . '/../../../../../logs',
             'cache' => [
                 'enabled' => false,
                 'response_ttl' => 3600,
--- a/src/core/Directus/Util/Installation/stubs/config.stub
+++ b/src/core/Directus/Util/Installation/stubs/config.stub
@@ -8,7 +8,7 @@ return [
 
     'settings' => [
         'logger' => [
-            'path' => __DIR__ . '/../logs',
+            'path' => '{{logs_path}}',
         ],
     ],
 
--- a/src/core/Directus/Services/ProjectService.php
+++ b/src/core/Directus/Services/ProjectService.php
@@ -46,6 +46,7 @@ class ProjectService extends AbstractService
 
             'timezone' => 'string',
             'locale' => 'string',
+            'logs_path' => 'string',
 
             'project_name' => 'string',
             'app_url' => 'string',


src/core/Directus/Util/Installation/InstallerUtils.php
```diff
@@ -894,6 +894,7 @@ private static function createConfigData(array $data)
             'feedback_token' => sha1(gmdate('U') . StringUtils::randomString(32, false)),
             'feedback_login' => true,
             'timezone' => get_default_timezone(),
+            'logs_path' => __DIR__ . '/../../../../../logs',
```
rijkvanzanten: Will this use the directory of the installer utils or the config folder?

src/core/Directus/Util/Installation/stubs/config.stub
```diff
@@ -8,7 +8,7 @@ return [
 
     'settings' => [
         'logger' => [
-            'path' => __DIR__ . '/../logs',
+            'path' => '{{logs_path}}',
```
rijkvanzanten: I think this might evaluate to `'__DIR__ . '/../logs''`

src/core/Directus/Util/Installation/InstallerUtils.php
bjgajjar: It will use the installer utils directory

src/core/Directus/Util/Installation/stubs/config.stub
bjgajjar: No! Here we'll have a dynamic variable.` '__DIR__ . '/../logs'` this can be considered as a default value.

src/core/Directus/Util/Installation/InstallerUtils.php
rijkvanzanten: It has to be the config directory though... 😬 otherwise the logs will be in /src/Directus/Util/Installation instead of /logs

src/core/Directus/Util/Installation/stubs/config.stub
rijkvanzanten: Oh right, it evaluates to the the result of __DIR__. Let's make sure __DIR__ is the right folder and we're all set!

src/core/Directus/Util/Installation/InstallerUtils.php
bjgajjar: I meant `__DIR__` contain the utils directory and for converting it I used '/../../../../..'

So, basically, your logs will be at same place :) (`api/logs` folder)

rijkvanzanten: Perfect!

PR #1204, opened 2019-08-19
Fix #1203 [Remove readonly condition for duration field of directus_files table]

### LINKED ISSUE #1203: Unable to edit file duration from File Library UI
When we uploaded a video file (MP4) via the File Libary, the "Duration" property of the file isn't set. Neither can it be edited afterwards.  I'm logged in as an admin.

The only way to change the value appears to be editing the database record directly. Is it expected behavior? I can't seem to find much documentation about the duration property other than it is an integer.

image

Thanks

bjgajjar:
This PR will resolve the warning of undefined index `data` too when uploading file in the `development` env.

### DIFF
--- a/migrations/upgrades/schemas/20190819070856_update_directus_fields_field.php
+++ b/migrations/upgrades/schemas/20190819070856_update_directus_fields_field.php
@@ -1,0 +1,22 @@
+<?php
+
+
+use Phinx\Migration\AbstractMigration;
+
+class UpdateDirectusFieldsField extends AbstractMigration
+{
+    public function up()
+    {
+        $this->execute(\Directus\phinx_update(
+            $this->getAdapter(),
+            'directus_fields',
+            [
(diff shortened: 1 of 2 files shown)


benhaynes:
This works as a fix, but it opens up a new issue: now users can change the `duration` of an audio/video file. Maybe this is a good thing for some, but to me it's like allowing them to change `filesize` (seems wrong). 

In my opinion we should be fetching the duration after upload and storing that value in the database. The field itself should still be `readonly` by default. I think we have another Feature Request to let users change the settings for Directus System Collections... so they could make that field editable if they wanted to.

shartley76:
Thanks for the update. Will be useful to override or type in these fields if necessary. I have a local fix using ffmpeg to get and set the attributes for audio and video. it's a bit rough at the moment, but I can pass this back once tested here

PR #1208, opened 2019-08-21
Fix directus/app#1735 [ENUM support]

### LINKED ISSUE #1735: "Required" fails on some fields
v8.5.4

Altering 
image

To active state and save results in (this is from two different fields where the "truncated" file is the current filename)

image

And the result is that the fields required state is unaltered.

bjgajjar:
Contains the changes of API for https://github.com/directus/app/issues/1735

### DIFF
--- a/src/core/Directus/Database/Schema/DataTypes.php
+++ b/src/core/Directus/Database/Schema/DataTypes.php
@@ -24,6 +24,7 @@ final class DataTypes
     const TYPE_SORT                 = 'sort';
     const TYPE_STATUS               = 'status';
     const TYPE_STRING               = 'string';
+    const TYPE_ENUM                 = 'enum';
     const TYPE_TRANSLATION          = 'translation';
     const TYPE_UUID                 = 'uuid';
     const TYPE_DATETIME_CREATED     = 'datetime_created';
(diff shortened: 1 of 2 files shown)


bjgajjar:
@hemratna - Agree with you. But right now I tried to provide the support for those who are using their DB directly. If we are going to give the official support to `ENUM` type[Includeing add/update/delete]; then I'll suggest adding another interface for `ENUM` would be better.

When the user tries to add/delete an item for `ENUM` type using the dropdown interface; it will be tedious to perform as a backend performance.

Let me know your thoughts here.

hemratna:
I am suggesting to have a separate interface to fully support the `enum` datatype.

The current fix will support `enum` in existing tables and convert them into `dropdown` interface, But admin can not add, update or delete the enum types.

I am suggesting to merge this PR so all the existing tables may utilize this partially developed feature and let's plan for full-featured `enum` interface.

@benhaynes @rijkvanzanten Please advice.

rijkvanzanten:
Enum's are tricky to implement in the field setup, as they're data options rely on setting the values in the `length` field.

rijkvanzanten:
We can't hack in the arbitrary name `options` for support for enum's. The `options` key is reserved for data coming from `directus_fields.options`.

PR #1210, opened 2019-08-24
Improve YouTube Embed Provider

VincentDatrier:
Hello everyone,

I noticed the YouTube Embed Provider did not provide a way to parse youtu.be URLs.

I took upon myself brushing up my PHP, and came up with this solution. By replacing the `strpos` by a `preg_match` (with a regex found on StackOverflow), youtu.be is now a valid URL for a YouTube Embed.

I also added some code to support this URL scheme in the `parseURL` function; if "youtu.be" is in the url, it parses the url, gets the path and trims the `/` from it. If it's not the case, it goes to the original function.

I hope this could be useful!

### DIFF
--- a/src/core/Directus/Embed/Provider/YoutubeProvider.php
+++ b/src/core/Directus/Embed/Provider/YoutubeProvider.php
@@ -16,7 +16,7 @@ class YoutubeProvider extends AbstractProvider
      */
     public function validateURL($url)
     {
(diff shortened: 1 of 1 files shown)


hemratna:
@VincentDatrier Thanks a lot for PR. 🎉

benhaynes:
Thank you @VincentDatrier!! 😄

PR #1213, opened 2019-08-26
added file meta data for audio/video

shartley76:
As per description from @benhaynes in #1204 this feature will dynamically set width,height and duration for video and duration for audio files.
It uses the Char0n\FFMpegPHP php library, so need to add composer reference to add the required library when installing.
Due to our requirements for the files to be stored on AWS s3 with a signed url, it saves the file temporarily locally to determine the metadata.

Tested with 194Mb mp4 video file, and 30mb audio file.

### DIFF
--- a/src/core/Directus/Filesystem/Files.php
+++ b/src/core/Directus/Filesystem/Files.php
@@ -1,7 +1,6 @@
 <?php
 
 namespace Directus\Filesystem;
-require_once '/var/www/html/vendor/autoload.php';
 
 use Char0n\FFMpegPHP\Movie;
 use Directus\Application\Application;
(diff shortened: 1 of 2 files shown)


src/core/Directus/Filesystem/Files.php
```diff
@@ -1,7 +1,9 @@
 <?php
 
 namespace Directus\Filesystem;
+require_once '/var/www/html/vendor/autoload.php';
```
hemratna: This path might not work for all the users.
Can we use the char0n/ffmpeg-php via composer as mentation https://github.com/char0n/ffmpeg-php#composer-installation?

hemratna:
@shartley76 Thanks a lot for PR. 
Please use composer to add the package.

shartley76:
All good, will submit a new PR with the composer changes as per suggestions

PR #1214, opened 2019-08-26
Feature/audio video upload

shartley76:
Added composer support for ffmpeg as per prior PR guidance.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
     "ext-mbstring": "*",
     "ext-openssl": "*",
     "ext-gd": "*",
-    "webonyx/graphql-php": "^0.13.0"
+    "webonyx/graphql-php": "^0.13.0",
+    "char0n/ffmpeg-php": "^3.0.0"
   },
   "require-dev": {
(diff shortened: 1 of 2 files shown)


hemratna:
@shartley76 Can you please verify that after making last changes, are we getting `width`, `height`, `duration`?

Also the `use Char0n\FFMpegPHP\Movie;` doesn't have any effect (Showing unused, If not require then please remove from imports).

On my local machine (Mac with MAMP, PHP73) I am getting `null` for `width`, `height`, `duration`.

shartley76:
Hi @hemratna I can confirm that I can upload an mp4 video and mp3 audio file and it does set the properties as described. However I am using the S3 adapter, which i believe means you get a stream of data back . Looking at the code I will re-test with normal file system as I believe this code needs to be added to that option too.

shartley76:
an update @hemratna I have tested with local file adapter, and it does also set the properties correctly.
I'm running via docker on a linux, and adding the ffmpeg via apk add in my Dockerfile.
Not sure if there's a missing OS install of ffmpeg that you need to do on your OS?

PR #1224, opened 2019-08-28
Fix issue while update item.

hemratna:


JbalTero:
@hemratna the updating of item is fixed, now the deleting of item has the same error.

JbalTero:
@hemratna to replicate the delete issue: do [this](https://drive.google.com/file/d/1a9pUGmwJ954Mz6knJDbi2KF1Wa11OoX1/view?usp=sharing)

The issue is: Deletion on collection with one-to-many relationship and the delete mode is delete relationship.

JbalTero:
@hemratna created the issue here: https://github.com/directus/api/issues/1229

hemratna:
#1229 is not due to this PR, It's an already in `develop` branch. So let's track the delete issue separately on #1229.

JbalTero:
Thanks

Issue #1249, opened 2019-09-03
Files.php saveData: height is set to width


guidoschmidt:
# Bug Report
in [Files.php#L326](https://github.com/directus/directus/blob/master/src/core/Directus/Filesystem/Files.php#L326) the height of the media is set to the width.

## Steps to Reproduce

1. Upload a video
2. Check its meta data
3. Height property is equal to the width

## Expected Behavior
Width and height should be set properly..

## Actual Behavior
Height is always equal to the width of the file. Should be an easy fix, just replace $width with $height variable.

## Technical Details

- OS: Windows 10 (Version 10.0.17134 Build 17134)
- Web Server: Apache/2.4.39 (Win64) OpenSSL/1.1.1c PHP/7.3.8 (XAMPP)
 - PHP Version: PHP Version 7.3.8
 - Database: MariaDB, 10.4.6
 - Install Method: git clone from the master branch

bjgajjar:
Fixed in https://github.com/directus/api/pull/1251

PR #1250, opened 2019-09-03
FIX/core/Directus/Files.php: return $height instead of $width

guidoschmidt:
Fixes [issues/1249](https://github.com/directus/api/issues/1249).
Had to change it anyway for a current project, so here's a PR 🙂 
Directus is super awesome btw. thanks so much for the work and effort you put into it.


benhaynes:
Thanks @guidoschmidt!! Really appreciate you taking the time to submit a PR for this. 😄

PR #1253, opened 2019-09-04
Fix #2017[Renaming `Public` role gives `unauthorized` error on API calls]

### LINKED ISSUE #2017: How to implement storage adapters other than S3
From directus created by [michchan](https://github.com/michchan) : directus/directus#2087  

The documentation (6.4.0) just mentioned how to implement S3
but is there any guide to implement others, say dropbox, one drive etc.?

bjgajjar:


src/core/Directus/Application/CoreServicesProvider.php
```diff
@@ -694,7 +695,7 @@ protected function getEmitter()
                 $acl = $container->get('acl');
                 $tableGateway = new BaseTableGateway(SchemaManager::COLLECTION_ROLES, $zendDb, $acl);
                 $row = $tableGateway->select(['id' => $roleId])->current();
-                if (strtolower($row->name) === 'public') {
+                if (strtolower($row->name) === Roles::PUBLIC) {
```
shealavington: Will this work? because before it was checking that `'public' === 'public'`, now it's checking that `'public' === 2`

src/core/Directus/Services/RolesService.php
```diff
@@ -157,7 +158,7 @@ public function canDelete($id, $fetchNew = false)
         }
 
         // TODO: RowGateWay should parse values against their column type
-        return !(!$group || $group->id == 1 || strtolower($group->name) === 'public');
+        return !(!$group || $group->id == 1 || strtolower($group->name) === Roles::PUBLIC);
     }
```
shealavington: Same thing here, before it seemed to be checking that `$group->name === 'public'` now it's checking that `$group->name === 2`, right? 

I would imagine something like `Roles::getName(Roles::PUBLIC)` — Not sure on that code but, that's the idea haha, or instead of using the name at all, we could just change the logic to use group id instead.

bjgajjar: Agree with you @shealavington 

If I am right then this condition was there to make sure that the user should not delete `Public` or `Admin` role. Earlier user can remove `Public` and `Admin` role using `API`. So this code was not worked as expected.

Reverted this PR and sending another one which adds this functionality.

Issue #1271, opened 2019-09-06
InvalidProjectException raised when `DIRECTUS_USE_ENV=1`


puco:
<!--
For security issues, please email support@directus.io directly.

1. Do not delete this template or the issue will be closed
2. Ensure you're using the latest version of Directus
3. Post to the correct repo:
    App:  https://github.com/directus/app/issues
    API:  https://github.com/directus/api/issues (YOU ARE HERE)
    Docs: https://github.com/directus/docs/issues
-->

# Bug Report

## Steps to Reproduce

1. Run API with `DIRECTUS_USE_ENV=1` and other env. variables as required for Docker
2. Do any operation through APP (e.g. create collection)

## Expected Behavior
Collection gets created without issues.

## Actual Behavior
Collection gets created but an error is raised (possibly when adding this to audit log)

## Other Context & Screenshots

Exception: 
```
api[_].ERROR: Directus\Config\Exception\InvalidProjectException: Current configuration context only supports default project name (_) in /var/www/src/helpers/app.php:95
api_1       | Stack trace:
api_1       | #0 /var/www/src/core/Directus/Util/DateTimeUtils.php(112): Directus\get_project_config                                                                                                                                                                                                                                              api_1       | directus/api#1 /var/www/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(2526): Directus\Util\DateTimeUtils::nowInTimezone
api_1       | directus/api#2 /var/www/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(440): Directus\Database\TableGateway\RelationalTableGateway->recordActivity
[...]

PR #1272, opened 2019-09-10
Project name fix

WoLfulus:
This fixes issue #1271. 

It was due to a call that doesn't set the proper project name. This PR makes it default to use "_" the project name value.

Please check if this change doesn't impact any other environments before merging, as I just tested locally in docker.


benhaynes:
Thanks @WoLfulus!

dario30186:
Thank you for the fix. When the new docker image will be avalable ?

Issue #507, opened 2018-10-22
Check all CLI modules


WellingGuzman:
Check all CLI modules and delete what's not in used anymore and what needs to be fixed.

benhaynes:
Isn't that an important one @WellingGuzman ? How difficult is this to fix?

WellingGuzman:
There are some commands not being used, some not well implemented and other broken. This is not a critical problem, but we need to clean all these unused/broken commands.

It shouldn't be difficult to fix this. More time should go into docs and list the one we need and remove the one we don't need anymore.

WellingGuzman:
Summary:
- Cache module was removed. it was an under development module made for Directus 6.
- `db:install` removed. Uses `install:database` instead.
- `install:migrate` removed. Use `install:database` instead.
- `log:prune` `-s` option removed. Never in used.

PR #1274, opened 2019-09-11
Add CLI cache:clear

JbalTero:


JbalTero:
Just found [this](https://github.com/directus/api/issues/507#issuecomment-450432962)

Why is the cache module removed? being able to clear cache on the fly is important.

benhaynes:
Hmm, not sure. It seems that the code was written for Directus 6 (a completely different codebase) so maybe it wasn't working/applicable any more?

shealavington:
I would agree Ben, the post does mention

>it was an under development module made for Directus 6.

bjgajjar:
Hey @benhaynes 

Checked this code. And it is working as expected. After executing `cache:clear` in CMD it is removing all the cache. So there's no harm to merge. Let me know what are the further steps we should do for this PR :)

benhaynes:
Sounds good to me! Thanks Binal...

rijkvanzanten:
I don't think there's any additional steps needed for this right? LGTM!

PR #1275, opened 2019-09-11
Clear cache after project update/upgrade

JbalTero:


WoLfulus:
@benhaynes 

Does upudate/upgrade causes problem with the cache? I don't think this should be done by default because on big installations this can cause a huge impact (cloud for example)

benhaynes:
I'm actually not sure. I guess if the schema changed the cache would need to be invalidated, but maybe we can have a flag for that? Thoughts @rijkvanzanten ?

WoLfulus:
If it's a cache of directus metadata that would make sense. Not for user content.

bjgajjar:
> Instead of automatically flushing the caches on migrate, I think we should look at adding a cache-flush endpoint so you can flush the caches manually when needed

Agree with you. https://github.com/directus/api/pull/1274 will clear the cache from the command line. Still, we need to add an `API` endpoint. 
Lets's close this PR and generate one `API` endpoint instead of removing cache automatically.

JbalTero:
@bjgajjar agree with you. closing this. Thanks everyone.

PR #1280, opened 2019-09-12
Session-based auth flow along with kill functionality

bjgajjar:


rijkvanzanten:
Couple of questions so far:

* Do we need both `access_token` and `PHPSESSID`?
* Can we rename `access_token` to `session`?
* Does the API re-set the cookie when using it? [1]
* Does the API auto-flush expired cookies from the user_sessions table?

src/core/Directus/Application/Http/Middleware/ResponseCacheMiddleware.php
```diff
@@ -83,13 +83,11 @@ public function __invoke(Request $request, Response $response, callable $next)
...
-                        $cookie = new Cookies();
-                        $cookie->set('session',['value' => $authorizationTokenObject['token'],'expires' =>$expiry->format(\DateTime::COOKIE),'path'=>'/','httponly' => true]);
-                        $response =  $response->withAddedHeader('Set-Cookie',$cookie->toHeaders());
-                    }else{
-                        $response =  $response->withoutHeader('Set-Cookie');
-                    }
+                    $cookie = new Cookies();
+                    $expiryAt = $userSession ? $expiry->format(\DateTime::COOKIE) : DateTimeUtils::now()->toString();
+                    $cookie->set('session',['value' => $authorizationTokenObject['token'],'expires' => $expiryAt ,'path'=>'/','httponly' => true]);
+                    $cookie->set('PHPSESSID',['expires' => DateTimeUtils::now()->toString()]);
```
rijkvanzanten: Do we need this @bjgajjar? Why do we have this? What does it do?

rijkvanzanten: `PHPSESSID` that is

bjgajjar: - We need this to update the `expired_at` time of `session` cookie in every `API` call.
- As you have recommended we don't need `PHPSESSID` as a cookie. So trying to remove it from the response.

rijkvanzanten:
Does this set the cookie `path` based on the project used?

We should double check if you can be logged into multiple projects on the same API at the same time

src/core/Directus/Application/Http/Middleware/ResponseCacheMiddleware.php
```diff
@@ -96,7 +96,8 @@ public function __invoke(Request $request, Response $response, callable $next)
         if(isset($userSession)){
             $userSessionService->update($userSession['id'],['token_expired_at' => $expiry->format('Y-m-d H:i:s')]);
         }
-       
+        $response = $response->withHeader('Access-Control-Allow-Origin', $request->getHeader('Origin'))
+        ->withHeader('Access-Control-Allow-Credentials', 'true');
```
rijkvanzanten: This should probably be fetched from the config file:

https://github.com/directus/api/blob/develop/config/api_sample.php#L116

and be set to `true` by default

PR #1295, opened 2019-09-18
Extend config schema to support PDO options.

scottmckenzie:
Extending the schema with this option allows me to connect to MySQL over a TLS connection.

### DIFF
--- a/config/api_sample.php
+++ b/config/api_sample.php
@@ -24,9 +24,11 @@ return [
         // When using unix socket to connect to the database the host attribute should be removed
         // 'socket' => '/var/lib/mysql/mysql.sock',
         'socket' => '',
-        'driver_options' => [
-            PDO::MYSQL_ATTR_SSL_CAPATH => '/etc/ssl/certs',
-        ]
+        // Connect over TLS by using the appropriate PDO_MySQL constants:
+        // https://www.php.net/manual/en/ref.pdo-mysql.php#pdo-mysql.constants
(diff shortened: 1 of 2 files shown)


config/api_sample.php
```diff
@@ -24,6 +24,9 @@
         // When using unix socket to connect to the database the host attribute should be removed
         // 'socket' => '/var/lib/mysql/mysql.sock',
         'socket' => '',
+        'driver_options' => [
```
WoLfulus: Please comment these lines as they might cause issues with most people that are not using ssl

src/core/Directus/Config/Schema/Schema.php
```diff
@@ -38,6 +38,7 @@ public static function get() {
                 new Value('engine', Types::STRING, 'InnoDB'),
                 new Value('charset', Types::STRING, 'utf8mb4'),
                 new Value('socket', Types::STRING, ''),
+                new Value('driver_options', Types::ARRAY, []),
```
WoLfulus: As this didn't exist before, please use "?" to completely remove the value from the final object if it's not set. I'm not sure if make [] the default value can cause problems.

config/api_sample.php
scottmckenzie: done

src/core/Directus/Config/Schema/Schema.php
scottmckenzie: done

PR #1299, opened 2019-09-20
Add Aliyun OSS settings to config schema

putrasattvika:
Aliyun OSS settings required by [FilesystemFactory](https://github.com/directus/api/blob/8a1187ff0b3559ab7281c4990d5329d47d8835dd/src/core/Directus/Filesystem/FilesystemFactory.php#L80) is missing from the schema.

Also added `proxy_downloads` to the schema (implemented in #860).


bjgajjar:
PR seems okay to me. @WoLfulus - I don't think it will create any issue for `Docker` but still kindly look into once.

putrasattvika:
> Do we already support aliyun in api code? If we do, we're fine with this, otherwise it doesn't make sense to add the settings that is not being used.

Yes, support for Aliyun OSS was implemented in #939.

PR #1332, opened 2019-10-01
Fix #1329 [Add new fields to / endpoint]

### LINKED ISSUE #1329: Add new fields to / endpoint
# Feature Request

For the new login page, we'd like to have additional fields for images we can use. Currently we have `project_name` and `project_logo`. We should add `project_icon`, `project_color`, and `project_image`.

image

bjgajjar:
- Added 2 fields `project_image` and `project_icon` in global settings.
- '/' endpoint will return all the project related items.

### DIFF
--- a/src/helpers/settings.php
+++ b/src/helpers/settings.php
@@ -156,14 +156,18 @@ if (!function_exists('get_project_info')) {
      */
     function get_project_info()
     {
-        $settings = get_directus_settings_by_keys(['project_name', 'logo']);
+        $settings = get_directus_settings_by_keys(['project_name', 'logo','color','project_icon','project_image']);
 
         if (array_get($settings, 'logo')) {
             $settings['logo'] = get_project_logo_data(array_get($settings, 'logo'));
         }
+        if (array_get($settings, 'project_image')) {
+            $settings['project_image'] = get_project_logo_data(array_get($settings, 'project_image'));
+        }
 
         array_rename($settings, [
             'logo' => 'project_logo',
+            'color' => 'project_color'
         ]);
 
         return $settings;
--- a/migrations/db/seeds/FieldsSeeder.php
+++ b/migrations/db/seeds/FieldsSeeder.php
@@ -1129,7 +1129,7 @@ class FieldsSeeder extends AbstractSeed
                 'type' => \Directus\Database\Schema\DataTypes::TYPE_STRING,
                 'interface' => 'dropdown',
                 'locked' => 1,
-                'width' => 'full',
+                'width' => 'half',
                 'note' => 'The file-system naming convention for uploads.',
                 'sort' => 14,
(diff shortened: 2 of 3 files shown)


benhaynes:
Did you also add `project_color`?

bjgajjar:
`color` is already there. Doesn't it?

benhaynes:
Hmm, all I know is I don't think it's returned by this endpoint.

src/helpers/settings.php
```diff
@@ -156,14 +156,18 @@ function get_trusted_proxies()
      */
     function get_project_info()
     {
-        $settings = get_directus_settings_by_keys(['project_name', 'logo']);
+        $settings = get_directus_settings_by_keys(['project_name', 'logo','color','project_icon','project_image']);
```
bjgajjar: @benhaynes  - Here you go! :) I added these 3 variables in '/' endpoint as well as `project_icon` and `project_image` did not exist in the settings so added them in the `directus_settings` table.

benhaynes:
Awesome. Is now a good time to write a migration to update those two settings (color and logo) to add the `project_` prefix? Is that a breaking change? Or should we just add new settings keys with the correct names and leave the old ones?

@rijkvanzanten ?

bjgajjar:
> Is now a good time to write a migration to update those two settings (color and logo) to add the `project_` prefix?

It should not affect for the `APP` but it will be breakable for those who are using standalone API.

benhaynes:
Right. I can't imagine anyone is using those outside the app, but who knows. Safest bet would be to create new fields and leave the old ones... but will let Rijk decide.

rijkvanzanten:
If we leave the old ones, we'll end up with duplicate rows in directus_settings 🤔

benhaynes:
Not really...

`color` — currently used, but we would stop using and could eventually remove from new installs
`logo` — currently used, but we would stop using and could eventually remove from new installs
`project_icon` — could be added for when we switch to the square "module" sidebar
`project_color` — could be added specifically for the login bg

rijkvanzanten:
Lets not go with the `project_` prefix then, as it will be very confusing to have fields for `color` and `project_color` right next to eachother

benhaynes:
But I was using this as a way to "cross fade" between the two settings. We need the legacy ones now, and will move to the new/standardized ones with the new designs. Once we switch we just won't install the old keys any more and they'll disappear. So we won't have two next to each other once we release.

Or we just rename the old ones. Honestly, at a certain point we need to focus on codebase standards instead of completely avoiding all (potentially) breaking changes. And I _really_ doubt anyone is using `color` in their project. And if someone is, we can add a big notice to the release notes.

rijkvanzanten:
Agreed. Lets rename `color` to `project_color` with a migration and be done with it :+1

Issue #1337, opened 2019-10-03
Error parsing environment variables


raftAtGit:
# Bug Report

## Steps to Reproduce

1.  Run the API providing these additional 2 environment variables:
KUBERNETES_PORT_443_TCP="doesnt matter"
KUBERNETES_PORT="doesnt matter"
2. Run the API
3. Make a request to API

## Expected Behavior

API Call return : `{"error":{"code":3,"message":"Unauthorized request"}}`
Command run without errors

## Actual Behavior
API call returns error:
```
<b>Fatal error</b>:  Uncaught Error: Only variables can be passed by reference in /var/www/src/core/Directus/Config/Context.php:39
Stack trace:
#0 /var/www/src/core/Directus/Config/Context.php(39): Directus\Config\Context::expand('tcp://raft-test...', Array, 'tcp://raft-test...')
#1 /var/www/src/core/Directus/Config/Context.php(39): Directus\Config\Context::expand(Array, Array, 'tcp://raft-test...')
#2 /var/www/src/core/Directus/Config/Context.php(71): Directus\Config\Context::expand(Array, Array, 'tcp://raft-test...')
#3 /var/www/src/core/Directus/Config/Context.php(85): Directus\Config\Context::from_map(Array)
#4 /var/www/src/web.php(30): Directus\Config\Context::from_env()
#5 /var/www/public/index.php(3): require('/var/www/src/we...')
#6 {main}
  thrown in <b>/var/www/src/core/Directus/Config/Context.php</b> on line <b>39</b><br />

```
## Other Context & Screenshots
Looks like this happens due to nested nature of exploded environment variables. 
Adding below lines to `Context::expand` method just before return statement fixes the issue.
```
	if (!is_array($target[$segment])) {
		$target[$segment] = [];
	}

```

## Technical Details
- Install Method: running Docker image directus/api:v2.6.0-apache

nachogarcia:
I encountered I encountered this same issue while deploying on kubernetes.

bjgajjar:
Fixed in https://github.com/directus/api/pull/1347

Radionz:
Hi, so what is what is container version that fix that issue ?

Regards.

davidraviv:
Hey @bjgajjar
I also have this case while installing directus/api:v2.6.0-apache from Docker Hub on Kubernetes.
When do you expect the next release to be issued?

davidraviv:
Hey again @bjgajjar,
[...]

PR #1347, opened 2019-10-11
Fix parsing env variables

nachogarcia:
Fixes https://github.com/directus/api/issues/1337


bjgajjar:
LGTM. We can merge, once @WoLfulus  check this for `Docker`.

Issue #1297, opened 2019-09-19
GraphQL Endpoint is Very Slow


wc-matteo:
Tested both on your demo api and on a dedicated vps freshly installed with the latest version (no docker). For a simple query (no joins, filters, etc.), the time is always >1s.

Is this a known issue?

![image](https://user-images.githubusercontent.com/10562635/65250394-ccd71080-daf5-11e9-8401-7e21be1c0867.png)

DevSnap:
I have the same problem.

With queries on the same collection, on top-level fields:
- REST endpoint takes about 300ms (in the worst case less than 1 second) 
- GraphQL endpoint takes 5-7 seconds.

Unfortunately I have no technical information to enter.
In the logs folder there is nothing related. Can I activate a more verbose log?

Tested on custom docker implementation on aws ec2.

hemratna:
Directus generate GraphQL schema on each request. To generate schema we need to traverse all the relationship. 

To make it faster, I am suggesting to use Cache on server-side as well as client-side.

For the server-side caching, Please take a look at https://docs.directus.io/advanced/api/configuration.html#config-file-options

benhaynes:
The schema shouldn't change often once in production mode (I still think we should have a system flag for when the schema is "locked"), I feel like the API should cache that and just fetch data. 

We should definitely look into what info we can fetch once and cache to speed things up.

DevSnap:
Why not trigger the graphql schema generation when there is a change to the structure?

benhaynes:
Well we aren't always aware of a schema change, since Directus uses "database mirroring" admins can technically work directly in the database (not always through the API). So we would have to occasionally check the schema for changes. 

But this is essentially what I mean by "locking" the schema with a flag so we know when it's safe to cache the schema.

Ref: https://github.com/directus/api/issues/21
Ref: https://github.com/directus/api/issues/22
Ref: https://github.com/directus/api/issues/381

DevSnap:
You're right!

Otherwise i tried to enable the caching option, but unexpectedly the cache only works for that exact graphql request, just change it slightly and those 7/8 seconds are back.

hemratna:
@benhaynes We can add the GraphQL schema caching once we have schema flag.

Bedotech:
Maybe i find a bug that prevents a collection to be cached in the same request, don't cache the whole schema but make GraphQL schema generation faster: !1354

Please try it and let me know if is ok!

bjgajjar:
[...]

PR #1354, opened 2019-10-17
Fix wrong chache naming.

Bedotech:
We have recently had some performance problem in GraphQL implementation in Directus, the main problem is described here: #1297 

Digging down in the code with the profiler I have noticed there are some methods that are call numbers grows exponentially when we increase fields.

In particular `\Directus\Database\Schema\SchemaManager::getCollection` search for entry with `collections.` prefix but the method `\Directus\Database\Schema\SchemaManager::addCollection` store it with `table` prefix.

This kind of cache works only on in the request context but gives really a huge benefit to the GraphQL schema building.

In our case, the improvement is 67% in time, and the memory footprint is ~4% less.

This fix doesn't change the caching behavior between requests, so if the database changes the changes are reflected in the schema, in the first request that hit the modified database.

### DIFF
--- a/src/core/Directus/Database/Schema/SchemaManager.php
+++ b/src/core/Directus/Database/Schema/SchemaManager.php
@@ -789,7 +789,7 @@ class SchemaManager
         // save the column into the data
         // @NOTE: this is the early implementation of cache
         // soon this will be change to cache
-        $this->data['tables'][$name] = $schema;
+        $this->data['collections'][$name] = $schema;
     }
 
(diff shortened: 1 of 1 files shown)


bjgajjar:
It's fixing the wrong name. @hemratna - can you please confirm whether it makes any difference in the performance of GraphQL or not?

bjgajjar:
Hey @ale-rinaldi! We really appreciate you taking the time to review this. Thank you so much :)

PR #1375, opened 2019-10-26
Fix Warning:  Illegal string offset 'field' in src/endpoints/Settings.php on line 89 when query param contains single=1

JbalTero:
The issue happens when you request to `/_/settings?single=1`

Enable `development` mode to replicate it

### DIFF
--- a/src/endpoints/Settings.php
+++ b/src/endpoints/Settings.php
@@ -8,7 +8,7 @@ use Directus\Application\Http\Response;
 use Directus\Application\Route;
 use Directus\Services\SettingsService;
 use Directus\Services\FilesServices;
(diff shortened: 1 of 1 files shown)


src/endpoints/Settings.php
```diff
@@ -74,38 +74,49 @@ public function all(Request $request, Response $response)
          * Get all the fields of settings table to check the interface
          *
          */
-
-        $fieldData = $service->findAllFields(
-            $request->getQueryParams()
-        );
+        $fieldData = $service->findAllFields(ArrayUtils::omit($request->getQueryParams(), 'single'));
```
bjgajjar: Throwing warning class `'Directus\Api\Routes\ArrayUtils'` not found

PR #1391, opened 2019-11-06
Fix the general settings schema upgrade 1329/1389

arboehme:
Consider already existing fields from seeder. Otherwise the following exception occured trying to rename the fields in the file `20191105063515_update_general_setting_variabe_name.php`:

`{"error":{"code":null,"message":"SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'directus_settings-project_foreground' for key 'idx_collection_field'"}}`


rijkvanzanten:
Thanks a lot @arboehme!

PR #1405, opened 2019-11-08
Authenticate endpoint add 2FA exception

bjgajjar:
`Authenticate` endpoint will throw an error if 2FA is not enabled instead of encoding the needs2FA into the token.


rijkvanzanten:
This was merged into develop manually

PR #1418, opened 2019-11-12
Fixed primary key fields Graphql definition

boris-burgos:
The check for primary key fields was wrong so they never are defined as ID. So the Int! definition raises a Graphql validation error on primary key fields:

`Interface field Node.id expects type ID but ###.id is type Int!.`


Owlree:
Looks good to me.

As for coding style, may I suggest we break the line into a proper _if_ statement?

PR #1420, opened 2019-11-12
Make the use if the telemetry service configurable

arboehme:
The new config option for the API to enable / disable the using of telemetry service  (see https://github.com/directus/docs/issues/167)

### DIFF
--- a/config/api_sample.php
+++ b/config/api_sample.php
@@ -10,6 +10,7 @@ return [
         'logger' => [
             'path' => __DIR__ . '/../logs',
         ],
+        'useTelemetry' => true
     ],
 
     'database' => [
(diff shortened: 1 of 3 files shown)


rijkvanzanten:
The config file for the application has been removed in v8. I've implemented the same functionality in 4ccfb981.

rijkvanzanten:
My bad! Mistakenly closed the API one instead of the app one.

bjgajjar:
@arboehme - Thanks for the PR. But we should use the `telemetry` flag from the DB instead of the `config` file. You can find that flag from the `directus_settings` table. 

It'll be great help if you can update this PR based on DB instead of configuration. :)

arboehme:
> It'll be great help if you can update this PR based on DB instead of configuration. :)

@bjgajjar Sorry, I have not so much time in last time.... but, I see you already changed my implementation accordingly. Thank you very much!

PR #1437, opened 2019-11-15
Update Thumbnail.php

eudora-fabia:
Add checking if the format is supported by the Imagick.

If there is no checking, errors occur because Imagick doesn't support it.

For example, using Imagick 3.4.4, for SVG format, it will error: `no decode delegate for this image format `' @ error/blob.c/BlobToImage/353` and in effect the SVG 's thumbnail will respond with HTTP error 5XX instead of 404

For the commit update: https://github.com/directus/api/pull/1437/commits/d6956c607dd30668991cf1de325581c0e1f09304

The intention is to prevent `/_/files` endpoint in returning a thumbnail for unsupported non-image formats. If we return thumbnails even though it is not supported, then errors like above occurs.


bjgajjar:
Hey @eudora-fabia - It seems like the error still exist after applying this PR. Let's communicate in the slack channel for a further discussion.

bjgajjar:
This PR will not work for the system where an extension `Imagick`  is not loaded. As we discussed in slack, can you please update the code?

bjgajjar:
@eudora-fabia - Is there any new update on this issue?

PR #1447, opened 2019-11-19
Fix #1429 [Convert current timezone to UTC timezone]

### LINKED ISSUE #1429: Activity datetime is incorrect
If you save an item (eg: on the demo), the activity is saying `4 hours ago`... which is incorrect. This has been an issue in the past, it needs to be resolved completely this time.

Screen Shot 2019-10-24 at 3 59 45 PM

bjgajjar:
Fixes #1429

### DIFF
--- a/src/core/Directus/Application/CoreServicesProvider.php
+++ b/src/core/Directus/Application/CoreServicesProvider.php
@@ -265,11 +265,11 @@ class CoreServicesProvider
 
 
                 if ($dateCreated = $collection->getDateCreatedField()) {
-                    $payload[$dateCreated->getName()] = DateTimeUtils::nowInTimezone()->toString();
+                    $payload[$dateCreated->getName()] = DateTimeUtils::nowInUTC()->toString();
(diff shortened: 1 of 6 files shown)


everyx:
After we save all datetime related fields as UTC time at API, we should also check the APP interface is convert and send the datetime related fields value correctly.

benhaynes:
@everyx — correct. The App also needs to save the correct datetimes... but that's a little different because it can/should be user configurable (if you want to save in UTC, local timezone, or something else), whereas the system times are hardcoded to UTC for consistency.

everyx:
> @everyx — correct. The App also needs to save the correct datetimes... but that's a little different because it can/should be user configurable (if you want to save in UTC, local timezone, or something else), whereas the system times are hardcoded to UTC for consistency.

Yes, for APP, we should treat all datetime related field have a default local timezone by the user configured, and the interface should convert all to UTC time or datetime string with timezone info when the send the data to API.

Issue #1395, opened 2019-11-06
/server/ping should ignore project


rijkvanzanten:
The API currently returns a 500 server error with

```
code: 22
message: "Unknown project: _"
```

Seeing that the /server/ping endpoint will only verify that the server is reachable, I think it should ignore the `X-Directus-Project` header completely.

(PS: This should probably be a 404 instead of a 503 error too)

bjgajjar:
Fixed in https://github.com/directus/api/pull/1416

rijkvanzanten:
See https://github.com/directus/api/issues/1395

It's still trying to use the current project in it's usage of CORSMiddleware

PR #1453, opened 2019-11-20
Stop migrations execution for server endpoints

bjgajjar:
`/ping` endpoint was executing the migrations if the `project-name` was given in the request.
This PR will stop the execution of migrations for all the `/server` related endpoints.


rijkvanzanten:
With this PR applied, I'm getting 

`Statement could not be executed (42S02 - 1146 - Table 'rijk.directus_settings' doesn't exist)`

which basically means that this PR re-opens https://github.com/directus/api/issues/1395

rijkvanzanten:
My issue is a bit of an edge case, so lets merge this in, reopen #1395, and refactor out the CORSMiddleware that's trying to read the project config from this endpoint.

PR #1464, opened 2019-11-23
Adding command line support for new project flow

bjgajjar:
This PR will generate the `super-admin-token` for the first project installation and return it to the console. After the first installation, for every other `install:config` expects `-a` as a super_admin_token; it's verify with the generated one then only the project config file will be generated.


hufon:
Writing a __api.json in the config.json directory would break docker installations as it would require a persistent volume just for this file... 
It could be great if we can "pass" this super admin password via environment variables.

bjgajjar:
> It could be great if we can "pass" this super admin password via environment variables too.

IMO, this is not required. 

And I am not sure - this will affects the `Docker` or not.

rijkvanzanten:
The docker setup only supports a single tenancy through env variables to begin with, so you'd never have to use the super admin password

rijkvanzanten:
@bjgajjar Can you add this to the docs as well?

PR #1466, opened 2019-11-24
Migration changes

bjgajjar:
This PR will; 
- Rename `db` to `install`
- Remove `schema` and `seeds` from `db` and `upgrades`
- Merge the `seeds` and `schemas` in `db`
- Remove all the older migration files from `upgrades`
- Add migration file to remove these utilities for those users who are going to upgrade their system.


bjgajjar:
1. As we discussed in slack, I removed the old migration files - if you still want to revert it; the I'll add it.
2. Yes - if you are considering the RC release for fresh installation; then it'll be broke.
3. I'll update it.

We already discussed all these things on the slack channel. And the things are build based on that only. Still, these are the recommended changes then just give me the confirmation - I'll do the needful changes. :)

rijkvanzanten:
I think we had some miscommunication regarding the older migrations, especially in the "clean up the repo vs compatibility" discussions. Lets put the old migrations back, so we can have people update from any v7 version instead of just v7.11.

Having the cleaned up folder structure already helps a lot in remediating confusion in what's what.

I can confirm that fresh installations seem to work again. I'm a bit confused by what you mean by 

> Yes - if you are considering the RC release for fresh installation; then it'll be broke - as the user must be in the v7 version [Exact the last before the RC release] for executing v8

PR #1467, opened 2019-11-24
Rework the thumbnailer

### LINKED ISSUE #813: Dynamic default action and quality for thumbnail information
# Feature Request
Make action and quality used in generating thumbnail information in file helper configurable or dynamic in some way.

Currently the thumbnail information added to the image data is by default "crop" with quality "good".
See:
https://github.com/directus/api/blob/master/src/helpers/file.php#L210

A simple approach would be to add setting options "Default thumbnail action" same for quality. That could be used as a 4th/5th parameter in the file helper.

A more sophisticated solution would be to read thumbnail settings and create thumbnails for all configured thumbnail actions and qualities. But maybe the current default would result in too many entries ( 2 actions * 4 qualities per size ).

Don't know if this would be a good place for a first contribution but I would be willing to give it a shot.

urvashithakar:
In this PR,

- New thumbnail flow as discussed in #813
- As discussed with @bjgajjar, added migrations in upgrades folder

@benhaynes  @rijkvanzanten 
Let me know the list of options that are supported for thumbnail creation so I will do code according to that.

Resolves #813


philleepflorence:
Will this PR allow the Admin to select the default thumbnail size and quality in `directus_settings`?
Currently, V8-RC2 is generating the thumbnail from the first size in `directus_settings.thumbnail_dimensions`, and using `good` as the quality. 

Or will it generate all the thumbnail sizes that are listed?

rijkvanzanten:
There's a split between system thumbnails that can't be changed, and a user-provided whitelist that can be used to add any dedicated sizes you want

philleepflorence:
Okay, thanks! 
I will await the next update to see if it changes the current behavior.

Currently with these sizes:
`200x200,320x320,480x480,640x640,960x960,1280x1280,1600x1600,1900x1900`
When you hit the files endpoint, it generates an array of 8 thumbnail rows but with the quality of `good` and action `crop`.
My hope is:
1. We can change the quality and action.
2. Choose the default thumbnail size or turn off the additional thumbnail rows.

Also if for some reason the list is as follows:
`1280x1280,1600x1600,1900x1900,200x200,320x320,480x480,640x640,960x960`
The thumbnail generated is 1280x1280.

rijkvanzanten:
This is exactly what's possible now 🙂 

Instead of setting what sizes are available, you're required to set what formats are allowed, based on an array of objects containing

```
(key)
width
height
quality
fit
```

philleepflorence:
Okay, so this is the way it is designed?

> Also if for some reason the list is as follows:
> thumbnail_dimensions: 1280x1280,1600x1600,1900x1900,200x200,320x320,480x480,640x640,960x960
> The thumbnail generated in uploads is 1280x1280.

And querying the files endpoint returns an array of thumbnails for each of the sizes in `thumbnail_dimensions` for each file?

PR #1484, opened 2019-11-26
Config updates

rijkvanzanten:
* Rename `api-sample.php` to `_example.php`
* Updated `/projects` to ignore `_` prepended files when returning the project keys
* Moved `settings.logger` to `logger`
* Moved `app.env` to `env`
* Removed `nowInTimezone()` function
* Removed `app.timezone` config
* Removed `settings` config
* Removed `feedback` config


migrations/upgrades/20191123064129_update_user_role_type.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+
+use Phinx\Migration\AbstractMigration;
+
+class UpdateUserRoleType extends AbstractMigration
+{
```
bjgajjar: Delete this file - I updated this in `migration_changes` branch

src/endpoints/Server.php
```diff
@@ -34,20 +34,21 @@ public function __invoke(Application $app)
...
             $projectNames = [];
-            if(empty($scannedDirectory)){
+            if (empty($scannedDirectory)) {
                 throw new NotInstalledException('This Directus instance has not been configured. Install via the Directus App (eg: /admin) or read more about configuration at: https://docs.directus.io/getting-started/installation.html#configure');
-            }else{
+            } else {
                 foreach($scannedDirectory as $fileName){
-                    if(!StringUtils::startsWith($fileName, 'private.')){
-                        $fileObject = explode(".",$fileName);
+                    if(!StringUtils::startsWith($fileName, 'private.') && !StringUtils::startsWith($fileName, '_')){
```
bjgajjar: What if the user registered `_` as their project? For that scenario; `projects` endpoint will return `[]` as a data.

rijkvanzanten: Good point. I guess we can update this with a check for length == 1

PR #1487, opened 2019-11-26
Don't raise the required error if the field has a default value

rijkvanzanten:


src/core/Directus/Services/AbstractService.php
```diff
@@ -232,6 +232,8 @@ protected function createConstraintFor($collectionName, array $fields = [], $ski
             $isStatusField = $field->isStatusType();
             if (!$isRequired && $isStatusField && $field->getDefaultValue() === null) {
                 $isRequired = true;
+            } else if ($isRequired === true && $field->getDefaultValue() !== null) {
```
rijkvanzanten: This is the only bit that changed. My editor automatically removes empty whitespace, I recommend y'all enable that setting too 😉

PR #1490, opened 2019-11-27
Fix #1462 [Authentication API return user object]

### LINKED ISSUE #1462: Should `auth/authenticate` endpoint return user data as well as the token?
# Feature Request - Where it came from?

Somehow new tokens take some time to become available.
My mobile app logs the user in via `auth/authenticate` endpoint. Token is returned successfully. Then app tries to get the user data via `users/me` right after the token response. But API says `item not found` with status code 404. Then I re-called getting user after the failing attempt  and it works.

Returning user data with the token is the most basic and swifty way. But I do not know if you have security concerns on this.

bjgajjar:
Fixes #1462

### DIFF
--- a/src/endpoints/Auth.php
+++ b/src/endpoints/Auth.php
@@ -76,7 +76,7 @@ class Auth extends Route
                 default :
                     $this->storeJwtSession($responseData['data']);
             }
-            unset($responseData['data']['user']);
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
@bjgajjar Why did you unset `last_page` and `last_access_on`? Those values are very useful for the app 🙂 It uses the `last_page` value on login to know where to send the user on login.

PR #1492, opened 2019-11-27
Fix #1475 [Activity IP Address must return localhost]

### LINKED ISSUE #1475: Activity IP Address
Right now we're showing the IP address as `::1`, which I assume is localhost. This looks weird and  doesn't provide much info. Let's instead save `localhost` for these local IP ranges (unless there's something more descriptive we can save).

Also, please make sure that activity is sorted by `Action On — DESC` by default. Mine seemed to be sorted by `id ASC` which is not very useful.

bjgajjar:
Fixes #1475


rijkvanzanten:
It's confusing if the API starts returning data differently without having a clear source of where it's coming from. Normally, a Directus datatype will cause the API to cast data in certain types, which means that you can trace where a certain value transform is happening. Next to that, we have interfaces that take the raw API output and display it nicer on the front end. This solution is neither of those. Now we have a regular text field that all of a sudden returns data differently than stored, because it's a specific field in a specific collection. 

v6 was filled with these kinds of random exceptions and filters, which turned out to be _incredibly_ hard to maintain over time. We worked really hard to make sure everything would use the same system for data casting (directus_fields) in v7, and would hate to see that effort slowly undone..

Can we instead consider:
* adding an ip address directus datatype so it can do the casting based on that, or;
* save the ip address as `localhost` at the point where the API is generating the activity row, or;
*  add a ip-address interface on the client side that will render the ::1 value as localhost?

@benhaynes

benhaynes:
I am fine with any of these three options, but option one (dedicated datatype) seems like overkill... so I think I'd prefer one of the other two. I defer to you two! @rijkvanzanten @bjgajjar

everyx:
In my opinion, api should keep it simple, just store data, and let client side to decide how to show this data. So the third is best KISS solution.

> add a ip-address interface on the client side that will render the ::1 value as localhost?

rijkvanzanten:
Closed in favor of handling this on the app side

PR #1499, opened 2019-11-29
Thumbnailer changes

urvashithakar:


rijkvanzanten:
As I mentioned to @bjgajjar on Slack, I already started on some of the cleanup of the thumbnailer on the [`thumbnailer-cleanup` branch](https://github.com/directus/api/compare/develop...thumbnailer-cleanup), but apparently those changes weren't used.

Can you make sure the changes in there are included in here as well?

migrations/install/20180220023208_create_files_table.php
```diff
@@ -169,6 +174,16 @@ public function change()
                 'width' => 'half',
                 'required' => 1
             ],
+            [
+                'collection' => 'directus_files',
+                'field' => 'filename_download',
+                'type' => \Directus\Database\Schema\DataTypes::TYPE_STRING,
+                'interface' => 'text-input',
+                'locked' => 1,
+                'readonly' => 1,
```
rijkvanzanten: This shouldn't be readonly. The user is able to change the `filename_download` field at will to control how the file will be saved on download

migrations/install/20180220023208_create_files_table.php
```diff
@@ -169,6 +174,16 @@ public function change()
...
+                'hidden_detail' => 1,
```
rijkvanzanten: Shouldn't be hidden

migrations/install/20180220023208_create_files_table.php
```diff
@@ -169,6 +174,16 @@ public function change()
...
+                'hidden_browse' => 1
```
rijkvanzanten: ... nor on the browse page

migrations/install/20180220023243_create_settings_table.php
```diff
@@ -230,6 +230,22 @@ public function change()
...
+                'field' => 'file_naming',
+                'type' => \Directus\Database\Schema\DataTypes::TYPE_STRING,
+                'interface' => 'dropdown',
+                'locked' => 1,
+                'width' => 'half',
+                'note' => 'File-system naming convention for uploads',
+                'sort' => 31,
+                'options' => json_encode([
+                    'choices' => [
+                        'uuid' => 'File Hash (Obfuscated)',
```
rijkvanzanten: Is this a UUID or a hash? If it's a UUID, lets not call it a Hash

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        if (!$filesTable->hasColumn('filename_disk')) {
+            $filesTable->renameColumn('filename', 'filename_disk');
+        }
+    
+        if (!$filesTable->hasColumn('filename_download')) {
+            $filesTable->addColumn('filename_download', 'string', [
+                'limit' => 255,
+                'null' => true,
```
rijkvanzanten: This field can't be null. You can't download a file with `null` being the filename

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,90 @@
...
+            ])->save();
+        }
+        if(!$this->checkFieldExist('directus_files', 'filename_download')){     
+            $fieldsTable->insert([
+                'collection' => 'directus_files',
+                'field' => 'filename_download',
+                'type' => 'string',
+                'interface'=>'text-input',
+                'locked' => 1,
+                'readonly' => 1,
```
rijkvanzanten: Should be editable by the user. Make sure to also update hidden_detail and hidden_browse just like the seeder

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,90 @@
...
+                'hidden_detail' => 1,
+                'hidden_browse' => 1
+            ])->save();
+        }
+        if($filesTable->hasColumn('id')){
+            $filesTable->changeColumn('id', 'integer');
```
rijkvanzanten: The column already is an `integer` seeing that we're reverting the change to string plus the fact that that was never released.

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        if($this->checkFieldExist('directus_files','id')){
+            $this->execute(\Directus\phinx_update(
+                $this->getAdapter(),
+                'directus_fields',
+                [
+                    'type' => 'integer'
```
rijkvanzanten: See above

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,90 @@
...
+                ],
+                ['collection' => 'directus_files', 'field' => 'id']
+            ));
+        }
+
+        // adds private_hash in directus_files collection for existing data
+        
+        if ($filesTable->hasColumn('private_hash')) {
```
rijkvanzanten: Does the private hash column exist at this point? I don't see any "add private_hash" column migration above this statement, so I'm thinking the v7 users won't get this.

migrations/upgrades/20191128063221_add_file_naming_in_directus_settings.php
```diff
@@ -0,0 +1,34 @@
...
+                'collection' => 'directus_settings',
+                'field' => 'file_naming',
+                'type' => 'string',
+                'interface'=>'dropdown',
+                'locked' => 1,
+                'width' => 'half',
+                'note' => 'File-system naming convention for uploads',
+                'options' => json_encode([
+                    'choices' => [
+                        'uuid' => 'File Hash (Obfuscated)',
```
rijkvanzanten: See UUID comment from the seeder

src/core/Directus/Application/CoreServicesProvider.php
```diff
@@ -344,9 +343,9 @@ protected function getEmitter()
 
                 $payload->replace($recordData);
                 $payload->remove('data');
+                $payload->remove('filename');
                 $payload->remove('html');
 
-                $payload->set('id', $data['id']);
                 $payload->set('private_hash', get_random_string());
```
rijkvanzanten: It doesn't feel right to me that this `/asset/` endpoint specific code is added in so deep down the abstraction layer in CoreServicesProvider. Lets move that to AssetsService

src/core/Directus/Database/TableGateway/BaseTableGateway.php
```diff
@@ -433,7 +434,7 @@ public function updateRecordByArray(array $recordData)
         $recordId = $recordData[$primaryKey];
         if ($collectionName === SchemaManager::COLLECTION_FILES) {
```
rijkvanzanten: Same as above, lets move this `/asset` endpoint specific code out of the deep BaseTableGateway abstraction layer.

src/core/Directus/Database/TableGateway/BaseTableGateway.php
```diff
@@ -461,13 +462,13 @@ public function updateRecordByArray(array $recordData)
 
         $replace = true;
         if ($collectionName === SchemaManager::COLLECTION_FILES && static::$container) {
```
rijkvanzanten: Same as above. This kind of exception really shouldn't be this deep down the abstraction layer.

We have both `FilesService.php` and `AssetsService.php` to handle everything that's specific to saving files and retrieving assets. This should move into there

src/core/Directus/Filesystem/Files.php
```diff
@@ -336,8 +336,9 @@ public function saveData($fileData, $fileName,$fileId,$replace = false)
 
         $response = [
             // The MIME type will be based on its extension, rather than its
```
rijkvanzanten:
```suggestion
            // The MIME type will be based on its extension, rather than its extension
```

src/core/Directus/Services/AssetService.php
```diff
@@ -143,15 +120,14 @@ public function getThumbnail($params)
             }
             $result['mimeType']=$this->getThumbnailMimeType($this->thumbnailDir,$this->fileName);
             $result['file']=$image;
-            $result['filename']=$this->fileName;
+            $result['fileNameDownlaod']=$this->fileNameDownlaod;
```
rijkvanzanten: Why is `$result['fileNameDownload']` in camelCase? It's `filename_download` everywhere else.

src/core/Directus/Services/AssetService.php
```diff
@@ -77,35 +82,14 @@ public function getAsset($fileHashId, array $params = [])
...
+            return $this->getThumbnail($params);
         }
-        else {
-           $this->fileName=$uploadedFileName;
-           try {
-               return $this->getThumbnail($params);
-           }
-           catch(Exception $e)
-           {
-               http_response_code(422);
```
rijkvanzanten: Thanks for cleaning this up

src/core/Directus/Services/AssetService.php
```diff
@@ -77,35 +82,14 @@ public function getAsset($fileHashId, array $params = [])
...
         $file = $result->current()->toArray();
 
-        $ext = pathinfo($file['filename'], PATHINFO_EXTENSION);
-        $uploadedFileName = $file['id'].'.'.$ext;
-
-        if(count($params) == 0) {
-           $url=get_file_root_url();
-           $img = $this->filesystem->read($uploadedFileName);
-           $result=[];
-           $result['mimeType']=Image::make($this->filesystem->read($uploadedFileName))->mime();
```
rijkvanzanten: I added some logic in here that would retrieve the real last modified date from disk: https://github.com/directus/api/blob/76709d5ebb63ac839256f20f68cdb5c9487882de/src/core/Directus/Services/AssetService.php#L88-L92

src/core/Directus/Services/AssetService.php
```diff
@@ -114,18 +98,11 @@ public function getThumbnail($params)
         $this->thumbnailParams=$params;
 
         $this->validateThumbnailParams($params);
-
+
```
rijkvanzanten: You should enable the "Remove trailing whitespace" setting in your editor to prevent empty changes like this

src/core/Directus/Services/AssetService.php
rijkvanzanten: Also you have a type in download; ~downlaod~ -> download

src/core/Directus/Services/AssetService.php
```diff
@@ -333,6 +309,11 @@ public function load () {
         $ext = pathinfo($this->fileName, PATHINFO_EXTENSION);
         if (Thumbnail::isNonImageFormatSupported($ext)) {
             $content = Thumbnail::createImageFromNonImage($content);
+            // For non supported file format make default thumbnail with jpeg extension
```
rijkvanzanten: This sounds dangerous. Where is this based on? 

If someone tries resizing a PDF, you'll use the JPEG extension and see what happens?

We can't assume that everything we don't know the type of is JPEG

src/endpoints/Assets.php
```diff
@@ -16,28 +14,25 @@ public function __invoke(Request $request, Response $response)
...
-            header('Content-type: ' . $response['mimeType']);
-            header("Pragma: cache");
-            header("Content-Disposition: filename=".$response['filename']);
-            header("Access-Control-Allow-Origin: *");
-            header("Access-Control-Allow-Methods: PUT, GET, POST, DELETE, OPTIONS");
-            header("Access-Control-Allow-Headers: Access-Control-Allow-Headers,Content-Type");
-            header('Last-Modified: ' . gmdate('D, d M Y H:i:s \G\M\T', time()));
+            $response->setHeader('Content-type',$asset['mimeType']);
+            $response->setHeader('Content-Disposition','filename='.$asset['fileNameDownlaod']);
+            $response->setHeader('Last-Modified',gmdate('D, d M Y H:i:s \G\M\T', time()));
```
rijkvanzanten: The `Last-Modified` header controls whether or not the browser will refresh its cache. By setting the header to the current time every request, the browser will never cache any thumbnail. Read the last modified date from disk and use that instead. (I already built that on the thumbnailer-cleanup branch)

migrations/upgrades/20191128055729_update_directus_files.php
bjgajjar: https://github.com/directus/api/pull/1467/files#diff-107ca5f85e330f110332a173f909048eR88

String is on `develop` branch - this code will change the datatype of those into `integer`

migrations/upgrades/20191128055729_update_directus_files.php
bjgajjar: https://github.com/directus/api/blob/develop/migrations/upgrades/20191123053252_update_directus_files_field.php#L164

src/core/Directus/Application/CoreServicesProvider.php
bjgajjar: This is the older structure - in which the hooks are defined to save file action. May I know what exactly you are expecting to be changed?

bjgajjar: May I have a brief on what should be replaced? As this hook is predefined [Which execute the specific files operation on item.update and item.create event] - PR just updating the code based on our new requirement.

src/core/Directus/Database/TableGateway/BaseTableGateway.php
bjgajjar: I would say the same for this too - this is an [older code](https://github.com/directus/api/blob/8d4473c7a1eb1129b24174d5307f845fafe473e3/src/core/Directus/Database/TableGateway/BaseTableGateway.php). Let me know - if we are supposed to clean it in this PR.

rijkvanzanten:
Closed in favor of #1506

PR #1506, opened 2019-12-02
Update the thumbnailer

rijkvanzanten:
* Adds `filename_disk` and `filename_download` to differentiate between the filename on disk, and the preferred filename when downloading the file
* Re-enabled the filename format option


migrations/install/20180220023243_create_settings_table.php
```diff
@@ -405,6 +421,10 @@ public function change()
               'key' => 'file_mimetype_whitelist',
               'value' => ''
             ],
+            [
+                'key' => 'file_naming',
+                'value' => ''
```
rijkvanzanten: This should default to UUID

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,122 @@
...
+                        'directus_files',
+                        [
+                            'private_hash' => get_random_string()
+                        ],
+                        ['id' => $value['id']]
+                    ));
+                }
+            }
+        }
+        // Convert string ID of directus_files into integer
```
rijkvanzanten: The current directus_files id is already an auto increment integer. This will override everybody's primary-key, which is what we're trying to prevent with these changes

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,122 @@
...
+                            ],
+                            ['id' => $value['id']]
+                        ));
+                    }
+                }
+            }
+        }
+
+        if($filesTable->hasColumn('id')){
+            $filesTable->changeColumn('id', 'integer',['identity' => true]);
```
rijkvanzanten: This is already the case

migrations/upgrades/20191128055729_update_directus_files.php
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        if($this->checkFieldExist('directus_files','id')){
+            $this->execute(\Directus\phinx_update(
+                $this->getAdapter(),
+                'directus_fields',
+                [
+                    'type' => 'integer'
```
rijkvanzanten: The type is already an integer

rijkvanzanten:
The content-disposition header isn't set correctly

rijkvanzanten:
The API returns `thumbnails` for non-image typed assets (like PDFs)

image

These will never work, as the asset isn't an image

rijkvanzanten:
Opening a non-image asset through the asset endpoint returns

```
{
  "error": {
    "code": null,
    "message": "Unable to init from given binary data."
  }
}
```

PR #1524, opened 2019-12-04
Fix addUpgradeMigrations

rijkvanzanten:
It was reading the wrong folder and therefore didn't run. It also didn't work with ENV mode


WoLfulus:
🙏

PR #1573, opened 2019-12-16
Fix #1556[Return null as foreign object when record is soft deleted]

### LINKED ISSUE #1556: Collection Count Showing Less Than Actual
The Item Count in the search input is off: shows `198` when there are 22,000+ items in collection. I assume this is because of the math... saying there are less than page count of items (likely due to API not returning soft-deleted items).

This can be seen on the `tickets` table of the Demo.

bjgajjar:
When the relational child data is soft-deleted the API was unset the parent key from the response object; instead of that; this PR will return `null` as a child object and not remove the key itself.

But we should add a condition to remove the relations reference when the data is deleted from DB. We can consider it as a feature request. 

Fixes #1556


rijkvanzanten:
@bjgajjar does this change the regular API output at all? Does it only apply to relational items>

bjgajjar:
> does this change the regular API output at all?
Yes.

> Does it only apply to relational items?
Yes.

Kindly draw my attention if you found any issue.

rijkvanzanten:
> > does this change the regular API output at all?

> Yes.

😐So it's a breaking change? _What_ changed?

bjgajjar:
> So it's a breaking change? What changed?

Hey No. It's not breaking change Rijk. I mentioned in the description. 

It'll return the `null` instead of not returning the key itself. That's the change which this PR will do.

Though it's only for relation data.

rijkvanzanten:
Ok good. Just confirming we don't alter the API input / output besides fixing the bug mentioned 🙂

rijkvanzanten:
LGTM

PR #1582, opened 2019-12-19
Fix #1563 [Remove PHPSESSID from cookie]

### LINKED ISSUE #1563: API requires PHP_SESSID to be present / correct
In bigger (enterprise) Directus installations, it's common to have multiple instances of Directus connected to the same database, with a load balancer in front to send people to the different instances based on usage. The Directus application relies on Cookies for its authentication, which means that the load balancer has to use session stickiness to make sure the API is able to read the cookie correctly. 

Lets say you have:

* 1 Database
* You have 3 instances of Directus connected to this database
* 1 Load balancer

Enabling session stickiness on the load balancer level will make every request go to the same instance, negating the usefulness of having a load balanced setup in the first place. For normal CRUD operations, this isn't the end of the world, but it does become a problem for files. Generating / retrieving thumbnails is currently very heavy (see https://github.com/directus/api/issues/1542), which can result in load problems normally taken care of by the load balanced setup.

The user sessions are currently stored in the database, but for some reason the API is not able to work without the same `PHP_SESSID` also being present. This requires you to enable session stickiness, which in turn causes the problem mentioned above.

@bjgajjar Seeing we have the `directus-{project}-session` cookie which is managed through the database, is there anything we can do to drop the reliance on `PHP_SESSID` so it doesn't matter to what server you connect?

bjgajjar:
@rijkvanzanten - Though I am not sure that this will resolve #1563; I need your help to test it in multiple servers.

As we discussed in the slack message; removing `PHPSESSID` from the cookie will resolve the issue. 

I am not able to test it as I don't have a system setup with multiple servers.

Can you please verify it?


batschz:
Any update here?

rijkvanzanten:
No. It's very hard to reproduce and therefore test if this PR actually helps or not. Are you experiencing #1563 on your side @batschz?

rijkvanzanten:
Closed in favor of https://github.com/directus/api/pull/1736

PR #1583, opened 2019-12-19
Fix #1579 [Error uploading file with comma in the name]

### LINKED ISSUE #1579: Error uploading file with comma in the name
If you upload a file with a comma in the name the request to the assets thumbnail will return a:

ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION

This stackoverflow article suggest adding a content disposition to the response header and putting quotes around the file name:

`Response.AddHeader("Content-Disposition", "attachment;filename=\"" + filename + "\"");`

https://stackoverflow.com/questions/13578428/duplicate-headers-received-from-server

BTW, thanks for simplifying the issue submitting process to be less formal.

bjgajjar:
Fixes #1579

### DIFF
--- a/src/endpoints/Assets.php
+++ b/src/endpoints/Assets.php
@@ -19,19 +19,16 @@ class Assets extends Route
             $request->getQueryParams()
         );
 
-        if(isset($asset['file']) && $asset['mimeType'])
-        {
-            $response->setHeader('Content-type',$asset['mimeType']);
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Doesn't this set the content disposition header on reading the file? How does that change anything for uploading files?

Issue #1567, opened 2019-12-13
Record Owner should be editable / assignable


tonydamage:
# Feature Request

Add editable item owner field. Owner field allows colabolator to access / edit items, which was not created by them, but explicitly assigned by them.

## What problem does this feature solve?

Sharing / limiting edit access to items with external collaborators without requirement for them to create item.

Currently there is no way to limit / assign access to specific item after it was created. The functionality is similar to "createdBy", but this field could be editable.

In our use-case we have proposed this workflow:
Intern with lowest visibility creates record (visible to him and us).
Admin creates additional records over time and needs way to share access to this records with concrete intern (other interns should not have access to this item).

## How do you think this should be implemented?

New type could be added (owner) and ACL could be extended to use this new type as owner.
This would affect mostly Schema, ACL and TableGateway classes.

The proposed behaviour:
- If collection does have `createdBy`, does not have `ownedBy` - behaves as now (visibility `mine` is determined by createdBy field
- If collection does have  `createdBy` and  `ownedBy`
  - On Record creation if `ownedBy` is empty, use creator as owner
  - `ownedBy` is editable field
  - Visibility `mine`  is only determined by `ownedBy` (open for discussion, could be also any of two, 
    but using only `ownedBy` is simpler, since current code assumes only one owner.

## Would you be willing to work on this?

I can implement changes and propose pull request.

rijkvanzanten:
Instead of relying on the default `created_by` field, you can create your own `owner` field and use the regular `user` interface with the `user_created` type. The `user` interface is editable

rijkvanzanten:
We should consider renaming `user_created` to `owner` though and making that the default instead of `user_created`. @benhaynes

benhaynes:
Yeah, that would work. I guess it's implicit that an owner would be )initially_ set by the creator. Whether it can be changed after that is more of a setting.

bjgajjar:
Fixed in https://github.com/directus/api/pull/1585

PR #1585, opened 2019-12-20
Rename User-Created Interface to Owner

bjgajjar:
This PR will rename `user_created` interface to `owner`  as discussed in https://github.com/directus/api/issues/1567

But it'll not make it editable. Please let me know if we should make them editable.

### DIFF
--- a/migrations/upgrades/20191220102232_update_user_created_to_owner.php
+++ b/migrations/upgrades/20191220102232_update_user_created_to_owner.php
@@ -18,10 +18,4 @@ class UpdateUserCreatedToOwner extends AbstractMigration
             ['type' =>  'user_created']
         ));
     }
-
-    public function checkFieldExist($collection, $field)
-    {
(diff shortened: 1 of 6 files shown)


migrations/upgrades/20191220102232_update_user_created_to_owner.php
```diff
@@ -0,0 +1,21 @@
...
+    {
+        // To change the type of all the field of user generated collection to owner
+        $this->execute(\Directus\phinx_update(
+            $this->getAdapter(),
+            'directus_fields',
+            [
+                'type' => \Directus\Database\Schema\DataTypes::TYPE_OWNER,
+                'interface' => 'owner'
+            ],
+            ['type' =>  'user_created']
```
rijkvanzanten: We should check for the user created interface specifically here. People can be using other interfaces with the user_created type, and we don't want to override those

manuel-84:
It would be very useful to change this interface to behave like "user" when read_only option is manually disabled.
Otherwise, if someone add a owner field using the "user" interface, the permissions system option "mine only" will not be available

PR #1598, opened 2019-12-24
GraphQL Fixes

kiwimonk:
1) The filename type was renamed to filename_disk recently in the database.
2) o2m returns a blank array in some cases without the modification.


rijkvanzanten:
@bjgajjar can you ask @hemratna on your end why we need the Directus core types in separate files in the GraphQL implementation in the first place? IMO this stuff should be read from the database as well, just like the user collections:

https://github.com/directus/api/blob/201eda741bb6a13767e824ced1e51bb375589df7/src/core/Directus/GraphQL/Type/Directus/DirectusFieldType.php

Otherwise this problem will occur again as soon as we change another core collection's structure

hemratna:
@rijkvanzanten 
There are two main reasons for having the Directus core types in separate files.

1. There are some fields like `thumbnail`, `folders` who doesn't have a direct entry in `directus_fields` and `directus_relationship` tables. So I have to define such a type in [Graphql](https://github.com/directus/api/blob/201eda741bb6a13767e824ced1e51bb375589df7/src/core/Directus/GraphQL/Type/Directus/DirectusFileThumbnailType.php) and it's a relationship with [other entities](https://github.com/directus/api/blob/201eda741bb6a13767e824ced1e51bb375589df7/src/core/Directus/GraphQL/Type/Directus/DirectusFileType.php#L39) 

2. Graphql is a strong type in nature. To build the GraphQL schema on each request is heavy in terms of MySQL query as well as PHP computing. 
I added the fixed schema (like Directus inbuilt tables and it's a relationship) in code, no need to compute the schema on each request.

> Otherwise this problem will occur again as soon as we change another core collection's structure

We need to update the GraphQL schema accordingly to changes in structure. It's timeconsuming but good for performance.

If needed, we can remove some of the types from PHP files and fetch from DB, but as per my understanding, we can not remove all the types from PHP files.

rijkvanzanten:
> There are some fields like thumbnail, folders who doesn't have a direct entry in directus_fields and directus_relationship tables. So I have to define such a type in Graphql and it's a relationship with other entities

We can address that and add it to the DB

> Graphql is a strong type in nature. To build the GraphQL schema on each request is heavy in terms of MySQL query as well as PHP computing.

Aren't we doing that for the users collections regardless? Can we cache this information? We could write it to memcache / file and only rebuild the generated graphql schema on creation / update of collections/fields based on a hook.

I would really like to prevent this information to be in two places at the same time. Keeping the DB updated is already hard enough to maintain as is. Having another place with the same information in a slightly different format is asking for trouble.

hemratna:
@rijkvanzanten Let's plan to move this to DB. I will guide @bjgajjar to start working on this.

rijkvanzanten:
Move what to the DB? 

Any thoughts on the second point? If fetching system collection info is too slow, how are you doing it for the non-system collections?

hemratna:
> Move what to the DB?
> 
Move the Directus core types from PHP files to DB, just like non-system collections.

> Any thoughts on the second point? If fetching system collection info is too slow, how are you doing it for the non-system collections?

We don't have any special process for non-system collections. Users can enable the Directus cache. 

------------------------------------------------------------

The reason to keep the Directus core types from PHP was, 
For example, to generate the whole GraphQL schema, it will take 4 seconds (1 for Directus core types + 3 for non-system collections)
By adding the schema of the Directus core types in PHP files I am trying to save 1 second.

rijkvanzanten:
That makes sense. We should have 1 system in place that works for everything though. In regards of Directus caching; I think we should have a separate cache for these GraphQL types; 4 seconds latency makes it useless.

src/core/Directus/GraphQL/FieldsConfig.php
```diff
@@ -77,6 +77,20 @@ public function getFields()
                     $fields[$v['field']] = Types::userCollection($collection);
                     break;
                 case 'o2m':
+                    $relation = $this->getRelation('o2m', $v['collection'], $v['field']);
```
rdsubhas: @rijkvanzanten Hi, thanks for the work :)

There is no such method called `getRelation` in this file 😱 

This change is causing massive issues in the current release, makes the whole GraphQL endpoint not load in any way, because of the missing method...

rdsubhas: Raised https://github.com/directus/api/issues/1795...

PR #1603, opened 2019-12-28
filename to filename_disk in filesystem delete fn

jacobrienstra:
Pretty self explanatory. Wasn't actually deleting, even though the app looked like it was


benhaynes:
Thank you @jacobrienstra — we'll take a look asap!

Issue #1104, opened 2019-07-10
Duplicate error codes


WoLfulus:
# Bug Report + Suggestion

While working on some PRs I came across error codes all around the repository. I did try to find free error codes, yet it's so easy to mess things up by introducing duplicated errors and not keeping the docs updated with all the possible values.

As discussed in slack, one possible solution is to replace numeric error codes with string constants instead, for example:

`0020` => `MISSING_STORAGE_CONFIG`

Also, having a single source of truth for those errors (a single php file defining them) would make it a lot easier to both keep docs updated and introducing new exceptions to the codebase.

hemratna:
After reading the discussion over the Slack, I agree with @WoLfulus that we need some way to document (list) the all the error code in once place in the API codebase. 

I face the issue while working on #1064 I had to throw two new exceptions, and docs on that stage showing the `0020` is the last exception code used. 

I found `0021` is used in some other place but it's not documented. So I create `0022`, `0023` code for my use case, and update the docs too.

But the `0022` is also used in some places. So if we have one truth of source for all the `error_code` this does not happen at all. 

The approach mentation by @WoLfulus sound great.

shealavington:
:+1:  I'm in agreement, one source with a list of constants sounds like the better alternative. Especially with multiple people working on Directus, as it'll make the `error code finding` easier.

PR #1609, opened 2020-01-01
Fix #1574 [Return warnings / notices in API when doing something that is impossible]

### LINKED ISSUE #1574: Return warnings / notices in API when doing something that is impossible
When you're trying to fetch relational data that you don't have access to, the API will return the primary key (m2o) or nothing at all (o2m). It would be nice if the API could return some notices about fields you're trying to access that you don't have access to.

In the following example:

`GET /items/articles/5?fields=id,title,author.*`

```json
{
  "data": {
    "id": 5,
    "title": "My Article",
    "author": 1
  }
}
```

we're trying to read `author` relationally, but the API doesn't return it. While this is is expected and correct behavior, it isn't the most developer friendly. If you don't know that the API works this way, it's very confusing why you don't get the data you requested.

I propose adding a new top level key for notices / logs / warnings of this nature:

```json
{
  "data": {
    "id": 5,
    "title": "My Article",
    "author": 1
  },
  "notices": [
    {
      "code": "WAR_NO_ACCESS",
      "message": "You don't have access to read `authors`"
    }
  ]
}
```

bjgajjar:
Fixes #1574

Return the notice if the relational [O2M/M2O] field doesn't have access to read the collection - only for the development environment.

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -14,7 +14,7 @@ use Directus\Database\SchemaService;
 use Directus\Exception\ErrorException;
 use Directus\Exception\UnprocessableEntityException;
 use Directus\Permissions\Exception\ForbiddenCollectionReadException;
-use Directus\Permissions\Exception\ForbiddenFieldReadException;
+use Directus\Exception\ErrorCodes;
 use Directus\Permissions\Exception\PermissionException;
 use Directus\Permissions\Exception\UnableFindOwnerItemsException;
 use Directus\Util\ArrayUtils;
@@ -33,6 +33,7 @@ class RelationalTableGateway extends BaseTableGateway
     const ACTIVITY_ENTRY_MODE_PARENT = 1;
     const ACTIVITY_ENTRY_MODE_CHILD = 2;
 
+    protected $exceptionMessages = [];
     protected $toManyCallStack = [];
 
     /**
@@ -919,6 +920,8 @@ class RelationalTableGateway extends BaseTableGateway
 
         $result['data'] = $data;
 
+        $result['messages'] = $this->exceptionMessages;
+
         return $result;
     }
 
@@ -1236,12 +1239,13 @@ class RelationalTableGateway extends BaseTableGateway
             foreach ($results as $index => &$item) {
                 $statusId = ArrayUtils::get($item, $statusField->getName());
                 $blacklist = $this->acl->getReadFieldBlacklist($this->table, $statusId);
+
                 $item = ArrayUtils::omit($item, $blacklist);
+
                 if (empty($item)) {
                     unset($results[$index]);
                 }
             }
-
             $results = array_values($results);
         }
 
@@ -2074,7 +2078,14 @@ class RelationalTableGateway extends BaseTableGateway
             }
 
             $relatedTableName = $alias->getRelationship()->getCollectionMany();
+
             if ($this->acl && !$this->acl->canReadOnce($relatedTableName)) {
+                $this->exceptionMessages[] = [
+                    'type' => 'warning',
+                    'message' => "Can't read `" . $relatedTableName . "`: read access to `" . $relatedTableName . "` collection denied",
+                    'code' => ErrorCodes::WARNING_ACCESS_DENIED,
+                    'fields' => [$alias->getName()],
+                ];
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
> only for the development environment.

Why? Isn't this important information regardless of what mode you're in?

Also, how did you implement it? What does the API return now?

bjgajjar:
> Why? Isn't this important information regardless of what mode you're in?

As @benhaynes [mentioned](https://github.com/directus/api/issues/1574#issuecomment) - I implemented this condition.

>Also, how did you implement it? What does the API return now?
```
{
  "data": {
    "id": 5,
    "title": "My Article",
    "author": 1
  },
  "messages": [
    "Reading items from author collection was denied"
  ]
}
```

Same as described [here](https://github.com/directus/api/issues/1574#issuecomment-566127742).

benhaynes:
Oh, to be clear... Rijk is the Technical Lead for Directus. My ideas are just questions, thoughts, and recommendations, but we should make sure Rijk agrees with things before implementing. 😬

rijkvanzanten:
Three points:

#### 1. It should be returned on production
The messages should definitely be returned in production mode as well, as its the only indicator why something returns `null`. The API is in production by default, and these messages are intended as a helper for (new) users of Directus that don't realize they have to add certain parameters.

#### 2. Use objects instead of strings
There's no way to programmatically read / translate these on the front-end. It would be more useful if they follow a similar structure as errors with a dedicated code for the message, so we can use that on the front-end. For example:

```json
{
  "messages": [
    {
      "type": "warning",
      "message": "Can't read fields from `author`: access denied",
      "code": "WARNING_ACCESS_DENIED",
      "fields": ["author"]
    }
  ]
}
```

`fields` in this example would be an additional field that returns the affected fields outside of a translated human readable string. This can be used to translate the `message` correctly on the front-end.

Ref https://github.com/directus/api/issues/1104 for the error code style

#### 3. Use more descriptive message

Seeing `Reading items from author collection was denied` out of the blue doesn't really give me all the information we need / have. Lets go with something more descriptive like:

```
Can't read `author`: read access to `authors` collection denied
```

bjgajjar:
@rijkvanzanten - Let me cleat another confusion here - This error message will be displayed for the relational data in the listing API. Am I Right?

rijkvanzanten:
For now, yeah. This can and most likely will be used in the future for other things as well

bjgajjar:
@rijkvanzanten - Updated the code as you recommended. Kindly review it :)

rijkvanzanten:
Right now you have a hardcoded string `"WARNING_ACCESS_DENIED"`. Should we move those to a global "error codes" file so we have a single source of truth for all possible errors moving forwards? @WoLfulus

bjgajjar:
@rijkvanzanten @WoLfulus - I added a global file for error code. Please verify it and if you okay with that then let me know so I can move all other error codes in this global file.

WoLfulus:
LGTM but my only concern is that if we should still display the author id value if it's denied? I understand this as a data leak, even if it's only the "id". Think that this for some reason can be an "email" for example if it's used as the external key.

I'd fail the whole request in that case instead of returning partial results. You're accessing a forbidden collection, so you're not doing what you're supposed to do - this in case of not having access to a collection. I might be missing some additional scenarios here though.

rijkvanzanten:
So you're saying instead of returning a warning that you're attempting to read a sub-item fail the whole request. That works for me too, as long as we have a clear error message stating why

WoLfulus:
Basically yes. I'm only thinking about the row-level access at this point. For example having an item that belongs to user X and an item that belongs to user Y. How should it behave if we have an intermediate collection that references both of those items? Kinda tricky

rijkvanzanten:
Exactly, that's the question at hand here.

Lets say:
* you have an instance with categories and articles
* the two are connected with a many to many categories_articles
* You have full read access to `categories`
* Articles is `mine` only

The question now is:

When trying to fetch all `categories` with `fields=*.*.*`, do we:

A) Error out with 401 Unauthorized + message "Not allowed to read from `articles`", or;
B) Return all articles that you _can_ read, return the primary key for the ones you're not allowed to, and add `messages` array saying:
"Reading article 15 was denied"

@benhaynes @WoLfulus @bjgajjar

benhaynes:
I think **B**. If we error out every time there is a little data you don't have access to, it will get annoying really quick. I think it's better to show what we can... as long as you have permissions to the main parent item(s) you're requesting.

PR #1626, opened 2020-01-08
Fix #1581[Access thumbnails using the original filename]

### LINKED ISSUE #1581: Access thumbnails using the original filename
As of Version 8 thumbnail urls contain the `private_hash`.
For website SEO purposes it would be better to be able to have the original filename (`filename_disk` or `filename_download`) in the url .
Can this be added back in?

bjgajjar:
Added a setting `thumbnail_url_pattern`[Private Hash | Filename Download]; which will use for defining the thumbnail URL 

Fixes #1581

### DIFF
--- a/migrations/install/20180220023243_create_settings_table.php
+++ b/migrations/install/20180220023243_create_settings_table.php
@@ -260,7 +260,7 @@ class CreateSettingsTable extends AbstractMigration
             ],
             [
                 'collection' => 'directus_settings',
-                'field' => 'thumbnail_url_pattern',
+                'field' => 'asset_url_naming',
                 'type' => \Directus\Database\Schema\DataTypes::TYPE_STRING,
                 'interface' => 'dropdown',
                 'locked' => 1,
@@ -448,7 +448,7 @@ class CreateSettingsTable extends AbstractMigration
                 'value' => 'uuid'
             ],
             [
-                'key' => 'thumbnail_url_pattern',
+                'key' => 'asset_url_naming',
                 'value' => 'private_hash'
             ],
             [
--- a/migrations/install/20180220023243_create_settings_table.php
+++ b/migrations/install/20180220023243_create_settings_table.php
@@ -270,7 +270,7 @@ class CreateSettingsTable extends AbstractMigration
                 'options' => json_encode([
                     'choices' => [
                         'private_hash' => 'Private Hash (Obfuscated)',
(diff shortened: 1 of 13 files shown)


rijkvanzanten:
@benhaynes any good ideas about a good name for this? `thumbnail_url_pattern` doesn't really work, as its for the `/asset` endpoint (and we're generally moving away from the name thumbnail).

For reference, this is the feature that lets you decide if you want to access files through `/asset/:private_hash` or `/asset/:filename_download`.

`asset_identifier` / `asset_key_field` / `asset_key_type` .. ?

benhaynes:
`middleware_naming`
`asset_middleware_naming`
`asset_url_naming`
`asset_middleware_url`
Any of those?

rijkvanzanten:
`asset_naming` sounds good to me

benhaynes:
Do you think it's confusing to not include something about this being the "middleware"? It's not the actual asset name, it's the private/middleware asset name.

rijkvanzanten:
Hmm yeah I see what you're saying. The real option for that is called `file_naming` by the way.

`file_naming` vs `asset_naming` would be the difference between how the file is stored on disk and how the file is retrieved through `/assets`

benhaynes:
I see —  then I guess we could explain the difference in the field notes (though I hate relying on those). A few more:

`file_access_naming`
`file_url_naming`
`asset_url_naming`
`asset_access_naming`

rijkvanzanten:
My vote goes to `asset_url_naming`

dobl1:
Hi,
Do you know when these changes will be published as docker version ?

Regards.

WoLfulus:
@dobl1 

Whenever we publish a directus release, we also publish a matching docker image to hub :)

dobl1:
@WoLfulus 
Yes I supposed, but do you have a date ?
Last version is from 11 days on docker hub.

benhaynes:
In the next day or two... Docker releases are tied to our actual releases.

dobl1:
Hello, I've use today the latest 8.5.3 pushed few hours ago.
It still replace the _ by - in the files names.

benhaynes:
Hey @dobl1 — can you open a ticket for that so we make sure it's properly tracked?

dobl1:
Done here https://github.com/directus/directus/issues/2563#issuecomment-582404187

dobl1:
So issue here : https://github.com/directus/api/issues/1730

PR #1631, opened 2020-01-10
Fix #1523 [Issue with TableGatewayFactory and ACL in Endpoints]

### LINKED ISSUE #1523: Issue with TableGatewayFactory and ACL in Endpoints
# Bug Report

When using the **TableGatewayFactory** on a read-only collection, when **ACL** is set to **false** (no updates are going to be performed) an exception is triggered.

Only when read and update permissions are set does it work using `ACL = $acl`.

## Steps to Reproduce

1. See Code and Response below.

## Expected Behavior

In an Endpoint, using **TableGatewayFactory**, one should be able to fetch items when permissions are set to read all. A token with read access should be able to read collections without ACL (breaking change from previous versions).

## Actual Behavior

An exception is thrown when ACL is set to false.

Sample Code:
```
use Directus\Application\Application;
use Directus\Database\TableGatewayFactory;

$container = Application::getInstance()->getContainer();
$dbConnection = $container->get('database');
$acl = $container->get('acl');

$tableGateway = TableGatewayFactory::create('<collection>', [
    'connection' => $dbConnection,
    'acl' => false
]);
```

Sample Error Response:
```
{
    "error": {
        "code": null,
        "message": "Call to a member function getStatusesOnReadFieldBlacklist() on null",
        "class": "Error",
        "file": "~/Sites/Directus/suite/src/core/Directus/Database/TableGateway/RelationalTableGateway.php",
        "line": 1803
    }
}
```

## Technical Details

- Device: Mac Desktop
- OS: MacOS 10.14.6
- Web Server: Apache - MAMP 5.5
- PHP Version: 7.2.21 - MAMP 5.5
- Database: MySQL 5.7.26 - MAMP 5.5
- Directus Release Version: 8.0.0-rc.2

bjgajjar:
Fixes #1523.

Conclusion; when `ACL` is set as `false` user will get the result; though the `public` permission is not set. As permissions depend on `ACL` only and if it's `false` then we don't have any access of `permission`.

@rijkvanzanten - Please correct me if I misunderstood anything.

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -1789,11 +1789,13 @@ class RelationalTableGateway extends BaseTableGateway
             //Logic for blacklisted fields
             $field = explode('.', $column);
             $field = array_shift($field);
-            $fieldReadBlackListDetails = $this->acl->getStatusesOnReadFieldBlacklist($this->getTable(), $field);
-            if (isset($fieldReadBlackListDetails['isReadBlackList']) && $fieldReadBlackListDetails['isReadBlackList']) {
-                throw new Exception\ForbiddenFieldAccessException($field);
-            } else if (isset($fieldReadBlackListDetails['statuses']) && !empty($fieldReadBlackListDetails['statuses'])) {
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
I.. have no idea what you're talking about.. 😬

bjgajjar:
Do you have an idea of `ACL`?

rijkvanzanten:
Nope!

PR #1675, opened 2020-01-21
Fix #1660 [SQL error when changing field to not required]

### LINKED ISSUE #1660: SQL error when changing field to not required
## Steps to Reproduce

1. Go to a collection
2. Click on a field that is flagged "required"
3. Remove the required and save

## Expected Behavior
The field should save and change from required to not required

## Actual Behavior
Error 500
`Statement could not be executed (42000 - 1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(16777215) Not Null' at line 1)`

- Device: DigitalOcean Server
- OS: Ubuntu Server 18.04
- Web Server: Apache/2.4.29
- PHP Version: 7.4
- Database: MySQL 8.0.18
- Directus Release Version:8.4.0
  - App Version: 8.4.0
  - API Version: 8.4.0

bjgajjar:
Fixes #1660

### DIFF
--- a/src/core/Directus/Exception/SQLException.php
+++ b/src/core/Directus/Exception/SQLException.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Directus\Exception;
+
+use Directus\Exception\UnprocessableEntityException;
+
+class SQLException extends UnprocessableEntityException
+{
+    const ERROR_CODE = 404;
+
+    public function __construct($message = '')
+    {
+        parent::__construct($message, static::ERROR_CODE);
+    }
+}
--- a/src/core/Directus/Services/TablesService.php
+++ b/src/core/Directus/Services/TablesService.php
@@ -1321,9 +1321,7 @@ class TablesService extends AbstractService
         $this->updateColumnsRelation($name, array_merge($toAdd, $toChange));
 
         foreach ($fields as $fieldData) {
-            if ($fieldData['type'] != DataTypes::TYPE_O2M) {
-                $schemaFactory->addNotNullConstraint($fieldData);
-            }
+            $schemaFactory->addNotNullConstraint($fieldData);
         }
 
         $hookEmitter->run('collection.update', [$name, $data]);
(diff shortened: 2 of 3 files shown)


src/core/Directus/Exception/SQLException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Directus\Exception;
+
+use Directus\Exception\UnprocessableEntityException;
+
+class SQLException extends UnprocessableEntityException
+{
+    const ERROR_CODE = 404;
```
rijkvanzanten: Maybe we should skip error code 404? Feels like that'll get easily confused with "Not found"

src/core/Directus/Services/TablesService.php
```diff
@@ -1321,9 +1321,7 @@ protected function updateTableSchema(Collection $collection, array $data)
         $this->updateColumnsRelation($name, array_merge($toAdd, $toChange));
 
         foreach ($fields as $fieldData) {
-            if ($fieldData['type'] != DataTypes::TYPE_O2M) {
```
rijkvanzanten: Isn't the O2M check important here? I remember the issue we had where we tried setting `not null` on a non-existing column.

src/core/Directus/Exception/SQLException.php
bjgajjar: We can use any but I preferred the sequence from [here](https://docs.directus.io/api/errors.html#schema-error-codes)

src/core/Directus/Services/TablesService.php
bjgajjar: Yes, we don't need this condition now as `O2M` is an alias type and based on [this](https://github.com/directus/api/blob/62c65dd96d0ba42c5e7f5e6c4a9d8212fbe7c098/src/core/Directus/Database/Schema/SchemaFactory.php#L103) code it'll return.

rijkvanzanten: Ok! :sligh

src/core/Directus/Exception/SQLException.php
rijkvanzanten: Yeah I agree 👍We should start using string based error codes in the next version like `ERR_UNPROCESSABLE_ENTITY` (@WoLfulus) to prevent these kind of issues as well

PR #1708, opened 2020-01-29
Fix #1640 [File not removed from disk after deletion]

### LINKED ISSUE #1640: File not removed from disk after deletion
Directus v8.3.1
When deleting a file via API or APP, the file is removed from database but not from the disk.

bjgajjar:
Fixes #1640 

Remove all variations of thumbnails from the system. 

@rijkvanzanten - User reported for both originals as well as thumbs. But originals are already getting removed at my end. Please check it once at your end too.


rijkvanzanten:
Yeah same here.

PR #1734, opened 2020-02-06
fix #1733: remove unnecessary exception re-throwing

### LINKED ISSUE #1733: Code cleanup: re-throwing exceptions
I found this code in multiple locations:

```php
 try {
            doSomething();
        } catch (\Exception $e) {
            throw new SQLException($e->getMessage());
        }
```

IMHO, it's a bad practice to re-throw an exception without doing anything useful. It's much better to let the original exception bubble up so that the stack trace in the log is full and shows the original location.

https://stackoverflow.com/questions/5551668/what-are-the-best-practices-for-catching-and-re-throwing-exceptions

dur41d:


src/endpoints/Fields.php
```diff
@@ -110,17 +106,13 @@ public function update(Request $request, Response $response)
...
-        try {
-
-            $responseData = $service->changeColumn(
-                $request->getAttribute('collection'),
-                $request->getAttribute('field'),
-                $request->getParsedBody(),
-                $request->getQueryParams()
-            );
-        } catch (\Exception $e) {
-            throw new SQLException($e->getMessage());
```
rijkvanzanten: I believe some of these changes are a little too loose. This is the only place this exception is transformed from a generic `Exception` to an instance of `SQLException` for example, which in turn controls the error code provided in the API. Could you verify @dur41d ?

dur41d: @rijkvanzanten I found that exceptions are being handled differently in different places. Sometimes it's re-thrown as is and sometimes it being converted into a different exception type.  With the lack of convention i think letting the exception bubble up will simplify code and debugging as the stack-trace is not circumvented. The API convent convert any unhanded exception to error code 500 or whatever as it sees fit and in one place.

rijkvanzanten: Yeah I agree, but changing that right now means that the error codes that are returned will change, which is technically a breaking change....

Refactoring all the errors to be properly thrown, and then handled by bubbling up is probably too much effort right now, but would you be able to leave the "double catch" for the ones that do the exception mapping to different types? At least that way it isn't a breaking change 🙂

rijkvanzanten:
@dur41d would you be able to leave the "double catch" for the ones that do the exception mapping to different types? At least that way it isn't a breaking change 🙂

PR #1751, opened 2020-02-09
Fixed config file creation validation

dwene:
Moving PR from `directus/directus` project: https://github.com/directus/directus/pull/2580

`InstallerUtils::ensureCanCreateConfig` was checking if `/var/directus` is writable instead of `/var/directus/config`.

### DIFF
--- a/src/core/Directus/Util/Installation/InstallerUtils.php
+++ b/src/core/Directus/Util/Installation/InstallerUtils.php
@@ -470,7 +470,7 @@ class InstallerUtils
         $input = ArrayUtils::omit($data, ['private']);
         $configPath = static::createConfigPathFromData($path, $input);
 
-        static::ensureDirectoryIsWritable($path);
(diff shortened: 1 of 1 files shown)


dwene:
> Directus also needs to be able to write to logs, and uploads. Having it just check for config might break if for people who didn't set up their permissions fully.
> 
> Also, the directus/directus repo is just a mirror for the built releases 🙂 Mind opening this PR on directus/api?

I think this function `ensureCanCreateConfig`  is specifically checking to see if the config can be created though right? While I agree that somewhere there needs to be checks for `logs` and `uploads`, doesn't seem like they would go here right?

WoLfulus:
LGTM. As for breaking user's installation, if permissions aren't setup properly they won't even be able to create it anyways, would they?

rijkvanzanten:
That's correct

PR #1752, opened 2020-02-10
Rework tooling organisation

jooola:


jooola:
@WoLfulus The sonar scanner makes the tests fails, I think the access token is not available on PR's.
Don't know how to fix though !

.gitignore
```diff
@@ -1,11 +1,97 @@
...
-/phpunit.xml
-Thumbs.db
-.DS_Store
+
+# Phpunit
 .phpunit.result.cache
+
+# Php-cs-fixer
 .php_cs.cache
-composer.lock
```
WoLfulus: We won't commit `composer.lock` since this is now a package and not a full laravel application.

composer.json
```diff
@@ -6,15 +6,10 @@
...
   ],
-  "authors": [
-    {
-      "name": "WoLfulus",
-      "email": "wolfulus@gmail.com"
-    }
-  ],
-  "bin": [
-    "./bin/directus"
-  ],
```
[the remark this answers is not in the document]
jooola: Woups, sorry this has been erased by accident.

.gitignore
jooola: Noted !

jooola:
> Any reason to move some files to the repository's root folder?

Some reasons on what changed in this PR :
- First we should avoid having naming conflicts with *Unix folders naming. The var folder is supposed to get application run-time data (such as logs/pictures/databases), not build/test output, the /build or /dist folder is more common for this purpose. The same goes for etc, which is supposed to handle software run-time configuration (such as webservers/php message broker configurations), not development tools configuration files.

- Having them at the root or the project gives a better visibility, on what tools is used to maintain this project and it is easier for newcomers/everyone to embrace the default. That being said, I can understand that having all those files in the root folders might seems like a mess, this is why I didn't used the fully default configuration file names in order to hide them on linux. I would prefer using the default names, but I though the main idea of the etc folder was to clean the root folder.
- You mentionned that it will not be a common app/project, but more like a library/package. I would argue that gives us even more freedom to move dev tools config files to the root folder since it will be installed using composer.
- Convention over configuration => development tools embed in IDE's use the default configuration paths to run there respective tasks, would be nice to embrace the default and avoid more configuration.
- Avoid unnecessary complexity: path need  to go to parent folder first. (`../`)

I felt like hiding them using dotfiles was the in-between that might be accepted, but I suggest to use the default configuration names.

jooola:
@WoLfulus Should I propose this PR to the new repo or do you think you don't want it ?

WoLfulus:
@jooola yes, if possible we should move these changes to api-next repository :)

I'll DM you to check some stuff.

Thanks!

Issue #1756, opened 2020-02-11
ID Auto-Increment breaks when I uncheck "Hidden on Browse"


medteck:
When I uncheck "Hidden on browse" for the ID field, I can't create Items in this collection from the API nor the UI (app).

I get a 422 error from the API and this is what I find in directus logs :
`ERROR: Directus\Validator\Exception\InvalidRequestException: id: This value should not be blank. in /var/www/directus/src/core/Directus/Services/AbstractService.php:186`

• Version of Directus : 8.5.3

Thanks!

Edit:
I still get the same error when I check it again. I had to delete the collection and create another one where "Hidden on browse" is checked to make it work.

hemratna:
@medteck 
I fixed this issue with PR #1764.
Please check it and if issue is solved then close it.

rijkvanzanten:
Fixed in https://github.com/directus/api/pull/1764

medteck:
Thanks guys!
But I can't pull the changes because my repository is directus/directus (suite).
Will you update it in the next versions ?

rijkvanzanten:
Yup! I'm planning on pushing out a release tomorrow afternoon (EST)

medteck:
Awesome! Keep up the great work 💯

Issue #1749, opened 2020-02-08
v8.5.4: update collection repeater field options get Directus\Exception\SQLException


everyx:
• Version of Directus: v8.5.4
• Error message (if any)
```shell
api_1             | [2020-02-08 12:27:18] api[xxxxxx].ERROR: Directus\Exception\SQLException: Statement could not be executed (42000 - 1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(16777215) Null' at line 1) in /var/www/html/src/endpoints/Fields.php:122
api_1             | Stack trace:
api_1             | #0 [internal function]: Directus\Api\Routes\Fields->update
api_1             | #1 /var/www/html/vendor/slim/slim/Slim/Handlers/Strategies/RequestResponse.php(40): call_user_func
api_1             | #2 /var/www/html/vendor/slim/slim/Slim/Route.php(281): Slim\Handlers\Strategies\RequestResponse->__invoke
api_1             | #3 /var/www/html/src/core/Directus/Application/Http/Middleware/AbstractRateLimitMiddleware.php(34): Slim\Route->__invoke
api_1             | #4 [internal function]: Directus\Application\Http\Middleware\AbstractRateLimitMiddleware->__invoke
api_1             | #5 /var/www/html/vendor/slim/slim/Slim/DeferredCallable.php(57): call_user_func_array
api_1             | #6 [internal function]: Slim\DeferredCallable->__invoke
api_1             | #7 /var/www/html/vendor/slim/slim/Slim/MiddlewareAwareTrait.php(70): call_user_func
api_1             | #8 /var/www/html/src/core/Directus/Application/Http/Middleware/AuthenticationMiddleware.php(124): Slim\Route->Slim\{closure}
api_1             | #9 [internal function]: Directus\Application\Http\Middleware\AuthenticationMiddleware->__invoke
api_1             | #10 /var/www/html/vendor/slim/slim/Slim/DeferredCallable.php(57): call_user_func_array
api_1             | #11 [internal function]: Slim\DeferredCallable->__invoke
api_1             | #12 /var/www/html/vendor/slim/slim/Slim/MiddlewareAwareTrait.php(70): call_user_func
api_1             | #13 /var/www/html/src/core/Directus/Application/Http/Middleware/TableGatewayMiddleware.php(25): Slim\Route->Slim\{closure}
api_1             | #14 [internal function]: Directus\Application\Http\Middleware\TableGatewayMiddleware->__invoke
[...]

Issue #1691, opened 2020-01-22
API removes auto increment if hidden on browse is changed


ithiru:
Version - v8.4.0
Error Message - id: This value should not be blank.

Steps

- Created new collection - turn on the Sort toggle, Modified On and Modified By
- Edit the Default primary key column 'id' by selecting it
- Go to schema Tab, Turn off 'Hidden on Browse'
- Click on 'Next'
- Click on 'Save'
- Go to the newly created collection to add new Row
- Click  on '+' at top right
- You will see only status with default 'Draft' selected - Leave it as is
- Click on 'Tick' button at top right.
- Displays Error Message 'id: This value should not be blank.'

My expectation is that the id column should be displayed on the list, but during add/edit, should not display. During add it should auto generate the id.

elfatherbrown:
Can confirm exact same case.

ithiru:
I took the tag 8.5.0 and tried this, I see the same issue. It is not allowing and throwing the same error. This is the steps I followed.

git pull
git checkout v8.5.0
php bin/directus db:upgrade -N - Verified the migrations ran by checking the directus_migrations table.

Accessed the app via url,
modified the id field and disable Hidden on Browse

Add new Item to collection and gets the same error. - 'id: This value should not be blank.'

Am I missing anything? Please assist

ithiru:
Tested the same in v8.5.4 and the default value is not saving..:-(

urvashithakar:
@rijkvanzanten  It is related to #1601

hemratna:
@rijkvanzanten I think it is related to #1601.

hemratna:
@ithiru 
I fixed this issue with PR #1764.
Please check it and if issue is solved then close it.

ithiru:
@hemratna Thanks, did this change go to directus/directus repo? I see the latest tag is v8.5.5. Should I wait the push or the 8.5.5. has all these changes?

Please confirm

rijkvanzanten:
@ithiru this hasn't been released to directus/directus yet. I'm planning on pushing out a release tomorrow.

rijkvanzanten:
Fixed in #1764

Issue #1668, opened 2020-01-15
Field default value not saving in Directus v8.4.0-v8.5.5


GMTA:
## Steps to Reproduce

1. Create a new collection
2. Add a text or status field
3. Set a default value
4. Save the field
5. Open the field again: the default value is gone

## Expected Behavior

The default value is stored in the table's schema.

## Actual Behavior

The default value is not stored in the table's schema.

## Technical Details

- Device: Desktop / Firefox 72.0.1
- OS: macOS Catalina 10.15.2
- Web Server: nginx 1.14
- PHP Version: 7.3
- Database: tested on MariaDB 10.2, 10.3 and MySQL 8.0.15
- Directus Release Version: 8.4.0
- Install Method: clone, checkout of v8.4.0 tag

Going back to 8.3.1 solves the issue.

mrfresh22:
Directus: 8.4.0
PHP: 7.4.1
MySQL: 5.7.28
I'm having troubles with this bug. When I edit "ID" primary key stops working

mrfresh22:
demo has the same bug
1. https://demo.directus.io/admin/#/thumper/settings/collections/services
2. Update Field: Status
3. Default Value set "published"
4. Saved
5. Check Default Value 
6. Omg? NULL?

mrfresh22:
2. Update Field: ID
3. Saved
4. https://demo.directus.io/admin/#/thumper/collections/services/+
5. "id: This value should not be blank."

AlbertoAbruzzo:
> > Please make sure to run this on PHP 7.3 and MySQL 8.0.15.
> 
> Ah, that's most likely what the problem is.. Directus doesn't support MySQL v8 yet..

@rijkvanzanten on version 8.3.1 worked fine with MySQL v8

GMTA:
> > Please make sure to run this on PHP 7.3 and MySQL 8.0.15.
> Ah, that's most likely what the problem is.. Directus doesn't support MySQL v8 yet..

Ah, based on "MySQL 5.7+" as stated on the page below, I thought it was supported.

https://docs.directus.io/installation/git.html

But it seems weird that the same database version worked in 8.3.1 but not in 8.4.0?

GMTA:
I've tested on MySQL 5.7 with Directus v8.4.0 and the same issue pops up. Might this be the same as #1676 ?

ithiru:
I took the tag 8.5.0 and tried this, I see the same issue. It is not persisting default value. This is the steps I followed.

[...]

PR #1764, opened 2020-02-12
Fix 1759 [Changing signed state doesn't work]

### LINKED ISSUE #1759: Changing signed state doesn't work
Sending `{ signed: false }` in a PATCH to /fields doesn't set the signed state to false for a numeric value and returns `{ signed: true }`.

hemratna:
fixes #1759

### DIFF
--- a/src/core/Directus/Database/Schema/SchemaFactory.php
+++ b/src/core/Directus/Database/Schema/SchemaFactory.php
@@ -95,33 +95,6 @@ class SchemaFactory
         return $table;
     }
 
-    // // Todo
-    // // Zend DB does not support NotNull constraint. This function will add/remove not null constraint from particular column
-    // public function addNotNullConstraint($column)
-    // {
-    //     // Make sure we don't add the constraint for alias type columns
-    //     if (DataTypes::isAliasType($column['type'])) {
-    //         return;
-    //     }
-
-    //     $connection = $this->schemaManager->getSource()->getConnection();
-    //     $sql = new Sql($connection);
-
-    //     $queryFormat = 'ALTER TABLE `%s`  MODIFY COLUMN `%s` %s';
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
How does this fix it? Have you tested this implementation? If so, how did you test it? 

This feels like a insecure solution, as this doesn't check for datatype of the column before attempting to set `signed`, or am I mistaken?

hemratna:
@rijkvanzanten  I want to mention one thing here. 
The following issues are related to #1601 
  #1765 
  #1759 
  #1749 
  #1739 
  #1756

So I am removing the #1601 code and trying to solve that PR related issue #1463. So after that all above mentioned issues will be solved.

hemratna:
After this commit It will solve all above mentioned issues

hemratna:
The PR #1601 for issue #1463 runs an `ALTER` query when the user modifies any column. 
The code for the `ALTER` query doesn't consider the current status of `autoincrement`, `unsigned`, `default`, etc. and alter for `required`.

In this PR I am removing the code of the PR #1601 and fix the issue #1463 by setting `$column->setNullable($nullable);` based on user input.

So by reverting PR #1601 all the above mentation issues will be fixed.

----------------------
Regarding the testing,

I created a collection with 
one textbox, set `required=false` as well as `required=true` (To test #1443)
one numeric field, set `required=false`, `required=true`, `signed=true`, `signed=false` (To test #1759 )

I also try to reproduce the below-listed issue after fixing this code in my local.

#1765
#1759
#1749
#1739
#1756
#1668 
#1691 

---------------------------

> This feels like a insecure solution, as this doesn't check for datatype of the column before attempting to set signed, or am I mistaken?

The code for the `signed` is already in place with default value @ https://github.com/directus/api/blob/master/src/core/Directus/Database/Schema/SchemaFactory.php#L216

But the PR #1601 override these values.

PR #1772, opened 2020-02-19
Fix #1735 ["Required" fails on some fields]

### LINKED ISSUE #1735: "Required" fails on some fields

hemratna:
This PR is dependent on #1764. Please merge the PR #1764 before this one.
Also, merge the doc PR https://github.com/directus/docs/pull/295

### DIFF
--- a/src/core/Directus/Database/Schema/SchemaFactory.php
+++ b/src/core/Directus/Database/Schema/SchemaFactory.php
@@ -48,6 +48,8 @@ use Zend\Db\Sql\Ddl\Constraint\PrimaryKey;
 use Zend\Db\Sql\Ddl\Constraint\UniqueKey;
 use Zend\Db\Sql\Ddl\CreateTable;
 use Zend\Db\Sql\Sql;
+use Zend\Db\Sql\Predicate\Expression;
+use Zend\Db\Sql\Select;
 use Directus\Database\Exception\FieldRequiredException;
 
 class SchemaFactory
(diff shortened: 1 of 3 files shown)


src/core/Directus/Database/Schema/SchemaFactory.php
```diff
@@ -154,6 +160,19 @@ public function alterTable($name, array $data)
         }
 
         $toChangeColumnsData = ArrayUtils::get($data, 'change', []);
+
+        // Throws an exception when trying to make the field required and there are items with no value for that field in collection
+        foreach ($toChangeColumnsData as $column) {
+            if($column['required']) {
+                $selectQuery = sprintf('SELECT * FROM `%s` WHERE `%s` IS NULL',$name,$column['field']);
```
WoLfulus: Let's change this to `SELECT COUNT` instead, otherwise it would fetch the whole table just to do this check.

src/core/Directus/Database/Schema/SchemaFactory.php
```diff
@@ -160,14 +160,14 @@ public function alterTable($name, array $data)
         }
 
         $toChangeColumnsData = ArrayUtils::get($data, 'change', []);
-
+        
         // Throws an exception when trying to make the field required and there are items with no value for that field in collection
         foreach ($toChangeColumnsData as $column) {
             if($column['required']) {
-                $selectQuery = sprintf('SELECT * FROM `%s` WHERE `%s` IS NULL',$name,$column['field']);
+                $selectQuery = sprintf('SELECT COUNT(*) as "count" FROM `%s` WHERE `%s` IS NULL',$name,$column['field']);
```
WoLfulus: Is there a reason why we're not using query builder here? I'm not sure where those parameters come from, and I'm not confident on merging this because of security issues.

Is it possible to change it to use a query builder instead?

PR #1775, opened 2020-02-20
Fix #1730 [ Prevented replacing '_'  by '-' in filename disk]

### LINKED ISSUE #1730: Use the filename to expose files
Hello,

Following of https://github.com/directus/directus/issues/2563, tested in latest 8.5.3.

The new filename_downloaded / filename_disk is causing some trouble as the filename is modified by directus and we do not have the choice to use the original filename to retrieve it through directus API.

So to be precise, we have two things :

- The filename-disk replace the `_` by `-` in the filename when importing a file.
- The filename-download is not used when choosing to use the File Name (Readable) for the links. (still using the filename-disk), No options to do that.

hemratna:
Fixes #1730

### DIFF
--- a/src/core/Directus/Filesystem/Files.php
+++ b/src/core/Directus/Filesystem/Files.php
@@ -624,7 +624,7 @@ class Files
     private function sanitizeName($fileName)
     {
         // Swap out Non "Letters" with a -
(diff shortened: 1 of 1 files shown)


hemratna:
@WoLfulus Can you please merge to master? I don't have the permission.

Issue #1729, opened 2020-02-05
Upload problems due to missing ffmpeg / ffprobe


maettyhawk:
Hi there,

I recently had problems uploading audio files (v8.4.0). The upload worked but the JSON response coming from the server was corrupt due to PHP warnings.

`<b>Notice</b>: Trying to get property 'format' of non-object in <b>/www/htdocs/dev/src/core/Directus/Filesystem/Files.php</b> on line <b>337</b><br /> <br /> <b>Notice</b>: Trying to get property 'duration' of non-object in <b>/www/htdocs/dev/src/core/Directus/Filesystem/Files.php</b> on line <b>337</b><br />`

I found out that ffmpeg / ffprobe is used to retrieve the format and the duration of the audios I uploaded, but on my shared hosting none of those is installed.
I wonder if it'd be possible to insert some conditions in PHP checking that ffprobe returns valid results instead of just assuming it like so:

`if( is_object($media) && $media->format && $media->format->duration){
                $duration = $media->format->duration;
 }`

Cheers

PR #1776, opened 2020-02-20
Fix #1542: Use file streaming instead of reading to memory

### LINKED ISSUE #1542: Use file streaming instead of reading to memory
Currently, the API reads file into memory on upload _and_ `/assets/` download, which makes it impossible to upload files that are bigger than ~100MB* and makes opening existing thumbnails slow.

We should use Streams to read / write / serve files, instead of reading them into memory whenever we can.

https://flysystem.thephpleague.com/docs/usage/filesystem-api/
https://odan.github.io/2018/03/09/emitting-chunked-stream-responses-with-slim.html

\* Depending on your server's memory and max php limits

alecreed:
Fixes #1542. Adds an optional boolean parameter $asStream (default false) to several methods on AssetService and makes use of that parameter in the Assets route. I verified locally with memory_get_peak_usage while serving the original of a large image that the file was no longer being loaded into memory in its entirety.

(This is my first pull request ever.)

### DIFF
--- a/src/core/Directus/Filesystem/Files.php
+++ b/src/core/Directus/Filesystem/Files.php
@@ -2,13 +2,12 @@
 
 namespace Directus\Filesystem;
 
-use Char0n\FFMpegPHP\Movie;
 use Directus\Application\Application;
+use Slim\Http\UploadedFile;
 use function Directus\filename_put_ext;
 use function Directus\generate_uuid4;
 use function Directus\is_a_url;
 use Directus\Util\ArrayUtils;
-use Directus\Util\DateTimeUtils;
 use Directus\Util\Formatting;
 use Directus\Util\MimeTypeUtils;
 
@@ -78,6 +77,7 @@ class Files
             $this->emitter->run('file.delete:after', [$file]);
         }
     }
+
     public function deleteThumb($file)
     {
         $ignoreableFiles = ['.DS_Store', '..', '.'];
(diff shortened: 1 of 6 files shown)


alecreed:
My mistake, #1542 also complains of files being fully loaded into memory on upload as well. Second commit also fixes that part. (Should also fix #1729 as well, since I was editing that method heavily and the relevant code stood out).

Basically, Filesystem\Files was making an on-disk copy of the temporary file created by php's built-in upload handling, then (for images at least) loading that copy into memory in order to call getimagesizefromstring on the actual bytes to extract metadata. I changed it to use the existing temporary file and to call getimagesize instead, which I believe only reads a minimal header from the file. Memory usage now looks stable regardless of the size of the uploaded file.

I only tested using the app and running on my local filesystem (not s3), so it's possible there are corner cases I'm not aware of.

everyx:
@alecreed If I'm free next week, I'll test s3 compatibility :)

rijkvanzanten:
This is super awesome. Thanks @alecreed! 

Lemme know what you find @everyx! I'd like to make sure we don't run into any unforeseen side effects when merging this 🙂

src/core/Directus/Filesystem/Files.php
```diff
@@ -406,13 +386,19 @@ public function saveEmbedData(array $fileInfo)
...
      * @param string $path - file path
      * @param bool $outside - if the $path is outside of the adapter root path.
+     * @param string|null $localCopy - path to local copy of the file, if available
      *
      * @throws \RuntimeException
      *
      * @return array file information
      */
-    public function getFileInfo($path, $outside = false)
+    public function getFileInfo($path, $outside = false, $localCopy = null)
```
WoLfulus: shouldn't this become a reference? &$localCopy since we check for $tmp on the `saveData` function?

Also, shouldn't we check for $tmp with `is_null` instead of isset?

alecreed: On the reference, I don't think so. It should always be a string path to a file (if it's not null), and it isn't modified inside getFileInfo or getFileInfoFromLocalCopy.

Do you mean `!is_null`? We only want to run the `ffprobe` if `$tmp` is a string path to a file (I think). I added that early on because previously it was possible for `$tmp` [to be undefined](https://github.com/directus/api/blob/5a21686fdb7b456ec8c8999ac8b3ea372c6a55e3/src/core/Directus/Filesystem/Files.php#L321-L326) at that point in the function. The fact it never seems to have caused runtime errors makes me suspect that `saveData` is currently only being used for file uploads (i.e. with an `UploadedFile` for the `$fileData` argument) even though it looks like it was written with an eye toward supporting a second option where `$fileData` would be a string of some sort (see some of the comments referencing base64 data, which I didn't change).

I'm new to the codebase, though, so I'm not sure if that's accurate and I wanted to err on the side of making fewer changes if I could.

everyx:
> This is super awesome. Thanks @alecreed!
> 
> Lemme know what you find @everyx! I'd like to make sure we don't run into any unforeseen side effects when merging this 🙂

Sure, so busy now, I'll test this weekend. :)

everyx:
@rijkvanzanten Just tested, it works fine with s3. :)

everyx:
@rijkvanzanten Can we merge this after @alecreed fix the conflicting?

rijkvanzanten:
I think so!

alecreed:
First conflict I've ever resolved. I think I did it right. It wasn't anything material to this change so much as an alternate way to fix a different issue. Should I resolve the @WoLfulus conversation or is that rude?

rijkvanzanten:
That's OK! As long as you don't resolve any open requests or comments without.. resolving them (either in code or in text) it isn't rude 🙂

src/core/Directus/Filesystem/Files.php
alecreed: I think we're good on this. Feel free to let me know of any other concerns.

rijkvanzanten:
I know it took a minute for me to review it, but @alecreed this is phenomenal! Thanks a lot ❤️

PR #1779, opened 2020-02-21
Fix 1729 [Upload problems due to missing ffmpeg / ffprobe]

### LINKED ISSUE #1729: Upload problems due to missing ffmpeg / ffprobe

hemratna:
Fixes #1729

### DIFF
--- a/src/core/Directus/Filesystem/Files.php
+++ b/src/core/Directus/Filesystem/Files.php
@@ -340,14 +340,19 @@ class Files
             $output = shell_exec("ffprobe {$tmp} -show_entries format=duration:stream=height,width -v quiet -of json");
             #echo($output);
             $media = json_decode($output);
-            $width = $media->streams[0]->width;
-            $height = $media->streams[0]->height;
-            $duration = $media->format->duration;   #seconds
+            if($media && is_object($media)) {
+                $width = $media->streams[0]->width;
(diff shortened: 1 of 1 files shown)


everyx:
Hi, @hemratna so we should install `ffprobe` as a dependency of directus? looks like it's missing in directus document.

hemratna:
I think `ffprobe` is an optional dependency. 
Ya, but it still a good idea to have things documented. 
I am open for PR. 🙏

PR #1782, opened 2020-02-24
fix wrong datatype length on text, mediumtext, longtext, tinytext

AleeeKoi:


AleeeKoi:
Already fixed by https://github.com/directus/api/pull/1764

PR #1793, opened 2020-03-03
Fix: GraphQL nested one to many

rijkvanzanten:
Original comment:

> The last patch fixed many to many relationships. The one to many relationships were still broken however. By checking the interface, the two relationship types can be handled differently.


rijkvanzanten:
These changes where already merged in #1598

PR #1798, opened 2020-03-05
Configure token TTL value

apiraino:
Hello Directus team,

I'd like the temporary token expiration to be configurable (currently defaults to 20 minutes). Currently it is not configurable.

This patch keeps the old behaviour, but allows the project config PHP file a new entry to configure the TTL token expiration, example:
```
    'auth' => [
        'secret_key' => 'XXX',
        'public_key' => '12345678-1234-1234-1234-123456789012',
        'ttl' => 60,
       .....
```
After a new Directus project installation, this behaviour can be controlled from the Directus backoffice panel:

Thanks for reviewing this!

cc: @herrfinke

### DIFF
--- a/migrations/upgrades/20200303111750_add_auth_ttl_field.php
+++ b/migrations/upgrades/20200303111750_add_auth_ttl_field.php
@@ -1,0 +1,69 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddAuthTtlField extends AbstractMigration
+{
+    public function up()
+    {
+        $settings = [
+            'auth_token_ttl' => [
+                // for directus_fields
+                'type' => 'integer',
+                'interface' => 'numeric',
+                'options' => '{\"iconRight\":\"timer\"}',
+                // 'locked' => 1,
+                // 'validation' => null,
+                'required' => 1,
+                // 'readonly' => 0,
+                // 'hidden_detail' => 0,
+                // 'hidden_browse' => 0,
+                'sort' => 25,
+                'width' => 'half',
+                // 'height' => null,
+                // 'group' => null,
+                'note' => 'How long the API authorization token will last',
+                // 'translation' => null,
+
+                // for directus_settings
+                'value' => 20,
+            ]
+        ];
+        foreach ($settings as $field => $options) {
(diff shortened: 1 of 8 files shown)


migrations/upgrades/20200303111750_add_auth_ttl_field.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddAuthTtlField extends AbstractMigration
+{
+    public function up()
```
apiraino: There's one thing I don't have clear. When I create a new Directus project, this migration is not applied (therefore I don't see the widget in the Directus backoffice. I tried also using the `change()` method to no avail.

Any hint on how to fix this?

reddo:
@apiraino, maybe specify the unit of measure in the field's description in the UI, as it is for the "Auto Sign Out"

apiraino:
@rijkvanzanten ping? :-)

migrations/upgrades/20200303111750_add_auth_ttl_field.php
rijkvanzanten: There is a folder of seeders too, you'll have to add the field there as well.

apiraino:
Hopefully we're there now :^)

apiraino:
@rijkvanzanten  ping?

WoLfulus:
LGTM. Is it playing well with app already?

apiraino:
> LGTM. Is it playing well with app already?

FWIW I've tested the changes by running fresh installs of Directus -> log in as admin -> change the Auth TTL value -> verify the changes in the DB.

rijkvanzanten:
The app relies on cookies and won't be affected by this change. We'll have to update the SDK though to ensure people relying on JWTs through the SDK don't run into problems

nathannau:
Warning !

This break the commande "line bin/directus user:password" !

THis return error : "ttl must be a number"

apiraino:
@nathannau Hi, good catch, thanks!

There is an error in how the new config item is retrieved, I've prepared a small patch to fix, I'm submitting a PR for evaluation.

PR #1808, opened 2020-03-16
compatible aliyun oss virtual hosted style

shouhuori:
Add aliyun oss and minio virtual hosted style support.Aliyun oss only support virtual hosted style，not allow use path style.

### DIFF
--- a/config/_example.php
+++ b/config/_example.php
@@ -75,6 +75,7 @@ return [
         //    'Cache-Control' => 'max-age=604800'
         // ],
         // 'endpoint' => 's3-endpoint',
+        // 'use_path_style_endpoint?'=> true,
 
         // Aliyun OSS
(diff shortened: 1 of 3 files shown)


shouhuori:
[reference](https://help.aliyun.com/document_detail/64919.html)

rijkvanzanten:
Thanks @shouhuori!

Is this meant to let Aliyun know that it should use the path in the endpoint? Unfortunately I can't read the reference so I'm shooting blind here :grin

shouhuori:
> 
> 
> Thanks @shouhuori!
> 
> Is this meant to let Aliyun know that it should use the path in the endpoint? Unfortunately I can't read the reference so I'm shooting blind here 😬

Aliyun Oss only support virtual hosted style and path style is deprecated.

PR #1822, opened 2020-03-24
Allow to set metadata for files on creation

dev7ch:
Attention, this it not tested, i hope this will work, due currently it looks like its not allowed to set metadata on creation, which would be nice


dev7ch:
Dear @bjgajjar do you see any obstacles on this?

dev7ch:
thanks 😊

inta:
`'metadata' => isset($file['metadata']) ? $file['metadata'] : null,`
Shouldn't that be called `$fileData` instead of `$file`?
Adding metadata is currently broken, but I don't know if this is the cause.

dev7ch:
mmh yes, could be .. i have not tested it

ThaDaVos:
This does definitely break metadata - also, this was not needed as one could already set `metadata` on file creation (we were doing this already)

@rijkvanzanten - can this PR be reverted and a `8.8.2` released?

rijkvanzanten:
> @rijkvanzanten - can this PR be reverted and a 8.8.2 released?

I'm afraid not. v8 has been end-of-life since November 2020, and won't receive any new updates from the core team.

PR #1843, opened 2020-04-07
Fix auth.ttl param retrieval

apiraino:
This should fix [this regression](https://github.com/directus/api/pull/1798#issuecomment-610238505). The new authorization time-to-live parameter was not correclty parsed when using the directus command line.

I've tested this change using the command line to change the password of a user

`$ bin/directus user:password --e user@host.com -p mypwd -k project_name`

and by creating a brand new project using the admin web app running on localhost (therefore project creation and migrations seem to work like before)


apiraino:
Hello :-) can this fix be pushed forward? Does it look ok? We would love to have this in the next minor Directus release

PR #1864, opened 2020-04-21
Relationship Limits and Statuses...

philleepflorence:
Limit the rows returned in relationships to the limit set in the parameters or use the system default limit.
Relationship status should follow the status set in the parameter.

**TODO**: Allow separate parameters for relationships...
E.G: relations.<collection>.limit or relations.<collection>.status.

### DIFF
--- a/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -35,6 +35,8 @@ class RelationalTableGateway extends BaseTableGateway
 
     protected $exceptionMessages = [];
     protected $toManyCallStack = [];
+    
+    protected $parsedParams = [];
 
     /**
      * @var array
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Lets come up with the ideal syntax for setting limits relationally so we can hit this out of the park in one go. How about `limit=200&limit[categories]=5&limit[authors]=2` ?

philleepflorence:
Using this will cause the first limit parameter to be dropped...

`limit=200&limit[categories]=5&limit[authors]=2`.

What about a separate variable to hold parameters for related collections?
This way it is backward compatible.

rijkvanzanten:
Mehhhh that sucks, would've been nice. 

I was thinking it would be nice if it matches the `fields` syntax in a way, but I can't really picture how that would be structured.. 🤔 

Something like

`&limit=500,categories.5,authors.2`

Not the nicest.. but at least it should work?

> What about a separate variable to hold parameters for related collections?

Like having the full string again?

```
?limit=500&sort=id&related[categories]=sort=title&limit=5
```

The above wouldn't really work with the "nested" query, so we'd have to overhaul that to have a different delimiter and assignment character. It gets pretty proprietary at that point 🤔 

```
?limit=500&sort=id&related[categories]=sort:title+limit:5
```

Alternatively, we could maybe do something like

```
?limit=500&sort=id&related[categories][sort]=5&related[categories][limit]=2
```

philleepflorence:
> ?limit=500&sort=id&related[categories][sort]=5&related[categories][limit]=2

I like this option!
It is keeping the same structure as the main collection parameter, so it should be easy for someone to add params for related collections...

rijkvanzanten:
It does get a little confusing with the parameters that already use a similar structure though

Currently, we allow for relational filtering through `filters[categories.title][contains]=book`, that wouldn't really be supported double in that idea above I suppose, otherwise you'd get... 

`related[categories[filter[title][contains]]=book` ?!

I don't know how I feel about that 🤢

philleepflorence:
I see your point...what about leaving filter the way it is, but also allow in the nested example?

We could pose the question within **Slack** or **Github Discussion** to get ideas?
I would imagine the solution needs to be backward compatible, correct?

philleepflorence:
I concur...

rijkvanzanten:
@WoLfulus @benhaynes any thoughts regarding the syntax for opening up the params to related fields? 

The goal here is to support things like limit, sort, etc on the related fields.

benhaynes:
To keep backwards compatibility, what about keeping everything the same as now but add new query strings for supported relational options? eg: `limit` _and_ `related_limit[field-scope]` ... Would this work?

Some naming options include:
* `limit_related[field-scope]`
* `limit_[field-scope]`
* `limit~[field-scope]`
* `limit.deep[field-scope]`

`?limit=200&related-limit[categories]=5&related-limit[authors]=2`
`?filters[categories.title][contains]=book&related-filters[authors][categories.title][contains]=book`

If we don't care about backwards compatibility, we could scope ALL query strings in an array, meaning each query string needs to have a `[]` after if for the parent, and scoped field name for related queries. The annoyance here is that 90% of queries have to add that extra set of brackets for the sake of consistency.

Thoughts @WoLfulus?

philleepflorence:
> If we don't care about backward compatibility, we could scope ALL query strings in an array, meaning each query string needs to have a [] after if for the parent, and scoped field name for related queries. The annoyance here is that 90% of queries have to add that extra set of brackets for the sake of consistency.

I favor the array option, so you can throw in different collections and fields in the array of objects. We can also make it backward compatible.

Imagine this for multiple collections:
```
limit[][collection-a]=20
limit[][collection-b]=300
status[][collection-a]=published
status[][collection-b]=draft,published
```
In the code, we can also allow for a single collection:
```
limit=200
status=published
```
This will translate to the main collection.

what do you think @benhaynes and @rijkvanzanten

benhaynes:
I defer to @rijkvanzanten on this one... it's all a bit technical for me, but I'm happy to review the final decision when we get there! :)

PR #1870, opened 2020-04-23
Dispatch hook actions when users attempt to sign into the Directus App.

philleepflorence:
HOOK: auth.success and auth.fail.
TODO: The directus_settings.login_attempts_allowed is not activated when authenticating through the App.

### DIFF
--- a/src/core/Directus/Authentication/Provider.php
+++ b/src/core/Directus/Authentication/Provider.php
@@ -176,7 +176,7 @@ class Provider
         // Verify that the user has an id (exists), it returns empty user object otherwise
         if (!password_verify($password, $user->get('password'))) {
 			
-			$hookEmitter->run('auth.fail', [$user]);
+			$hookEmitter->run('auth.fail', [$user, 'app']);
             $this->recordActivityAndCheckLoginAttempt($user);
             throw new InvalidUserCredentialsException();
         }
@@ -197,7 +197,7 @@ class Provider
 
         $this->user = $user;
         
-        $hookEmitter->run('auth.success', [$user]);
+        $hookEmitter->run('auth.success', [$user, 'app']);
 
         return $user;
     }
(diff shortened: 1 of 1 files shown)


philleepflorence:
@rijkvanzanten sure thing.
I need to add another parameter to differentiate between API Auth (which happens on all API Rest Calls and App Auth (which happens only when authenticating via the app).

rijkvanzanten:
I'm still wondering how much that split really exists. I've been having similar thoughts regarding caching. Ideally, we want to prevent caching coming in from the admin app, but you can also make the point that the admin app is just another project that connects to the API. I have been preventing the app from getting any preferential treatment, which also resulted in having an API that has the ability to do literally anything the app can. 

We could consider adding a "skip-hook" flag or something along those lines, but I'd rather not have any special treatment just because a request came from `/admin`

philleepflorence:
I like having the `no-cache` and `no-hook` flags!
Give control to the requesting application.

rijkvanzanten:
I'm actually about to take a swing at a `skip-cache` query param 🙂 Not the perfect solution, but at least it should make caching useful by being able to ignore the app in the caches

philleepflorence:
That is perfect for now...or using a `cache` parameter so caching is only done on request?

How would you add the `skip-cache` param to Admin Requests?

rijkvanzanten:
I think enabling caching should start caching the whole API automatically, with an escape hatch to circumvent it when absolutely needed (like when you're actively updating the contents). A skip-cache param should do the trick.

As for implementation in the app itself, I would update the JS SDK to have a "skipCache" option during initialization, and set that to `true` in the Directus admin app api.js file 🙂

philleepflorence:
That works!
Hopefully, that should fix the issue with the App not working with cache enabled and a user logs out.

rijkvanzanten:
Exactly! It doesn't fix all the cache related issues, but at least it takes care of the biggest one

PR #1876, opened 2020-04-30
Allow admin to manage password hash directely

nathannau:
Allow administrateur to get and set password hash directely on other user.
It will usefull to allow sync user between many project.

### DIFF
--- a/src/core/Directus/Application/CoreServicesProvider.php
+++ b/src/core/Directus/Application/CoreServicesProvider.php
@@ -498,7 +498,7 @@ class CoreServicesProvider
                 $rows = $payload->getData();
                 $userId = $acl->getUserId();
                 foreach ($rows as &$row) {
-                    $omit = [
+                    $omit = $acl->isAdmin() ? [] : [
                         'password'
                     ];
                     // Authenticated user can see their private info
@@ -553,6 +553,10 @@ class CoreServicesProvider
 
                     $type = $field->getType();
                     if (DataTypes::isHashType($type)) {
+                        if (is_array($value) && isset($value['hashed'])) {
+                            $payload->set($key, $value['hashed']);
+                            continue;
+                        }
                         $options = $field->getOptions() ?: ['hasher' => 'core'];
                         $hashedString = $hashManager->hash($value, $options);
                         $payload->set($key, $hashedString);
@@ -600,7 +604,6 @@ class CoreServicesProvider
                         $payload->set($key, $dateTime->toString(DateTimeUtils::DEFAULT_TIME_FORMAT));
                     }
                 }
-
                 return $payload;
             };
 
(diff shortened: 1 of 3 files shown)


src/core/Directus/Application/CoreServicesProvider.php
```diff
@@ -553,6 +553,10 @@ protected function getEmitter()
 
                     $type = $field->getType();
                     if (DataTypes::isHashType($type)) {
+                        if (is_array($value) && isset($value['hashed'])) {
+                            $payload->set($key, $value['hashed']);
+                            continue;
+                        }
```
rijkvanzanten, line 559: This'll break regular `hash` type fields. They way they work right now is to hash the raw value on creation. By not-hashing it when a value already exists, we'll effectively break hash-type support.

src/core/Directus/Database/TableGateway/BaseTableGateway.php
```diff
@@ -1761,6 +1761,7 @@ protected function validateRecordArray(array $record)
                 ($field && is_array($columnValue)
                     && (!DataTypes::isJson($field->getType())
                         && !DataTypes::isArray($field->getType())
+                        && !(DataTypes::isHashType($field->getType()) && isset($columnValue['hashed']))
```
rijkvanzanten, line 1764: Is this checking for the value of the column being the string `hashed`?

src/endpoints/Users.php
```diff
@@ -70,6 +70,13 @@ public function create(Request $request, Response $response)
             return $this->batch($request, $response);
         }
         $service = new UsersService($this->container);
+        $acl = $this->container->get('acl');
+
+        if ($acl->isAdmin() && isset($payload['password_hashed'])) {
+            $payload['password'] = [ 'hashed' => $payload['password_hashed']];
+            unset($payload['password_hashed']);
+        }
```
rijkvanzanten, line 78: I'd rather not add in a non-standard field like this 🤔

rijkvanzanten (changes requested):
You're looking to be able to update hash values without the API re-hashing the existing hash. However, by changing the hash to not run if a value is provided, you're effectively breaking the hash type fields altogether.. 

There is no real solution to this, as we can't have it both hash, and not-hash at the same time..

I think the only way forwards would be to introduce some sort of flag (in a query param?) that indicates that you want to skip the hashing step, because you're already providing a hash yourself.

rijkvanzanten:
Returning the `password` field for admins makes sense to me, but is only a small piece of the puzzle here 🙂

nathannau:
> 
> 
> Returning the `password` field for admins makes sense to me, but is only a small piece of the puzzle here 🙂

The rest is being read, but I'm running out of time

PR #1350, opened 2019-10-14
Feat/azure storage adapter

retrocede:
Add an option for azure file storage blobs as your storage provider.

PR #1892, opened 2020-05-15
Add azure blob storage as an option

retrocede:
Add an option for azure file storage blobs as your storage provider.

Re-implementation of  #1350 to resolve merge conflicts with master.


rijkvanzanten:
@retrocede Thanks! ✨ 

Can you add it to the recommend libraries here too?

https://github.com/directus/api/blob/db6c6f03083502e54da0bbe9a41d7a48250adaa6/composer.json#L47-L51

PR #899, opened 2019-04-19
Issue #886

itsmerhp:

PR #1938, opened 2020-06-20
Improve string to int/float cast

apiraino:
Hi,

I think there is a problem when the id of a record is a BIGINT. It looks like the string sent from the app is not always properly casted to integer/float resulting sometimes in a broken SQL query.

Example, we have this primary key `8004372939997067670` (a BIGINT in MySQL) and the generated query looks like (shortened for readability):

```
SELECT
    "test_table"."id" AS "id", ......
FROM
    "test_table"
WHERE
    CAST(`test_table`.`id` as CHAR) IN ('8.0043729399971E+18')
ORDER BY ...
LIMIT '1'
OFFSET '0'
```

To give more context on my research, from what I could understand, the cause of the issue is a brittle parsing introduced two years ago and not really fixed in a [subsequent patch](https://github.com/directus/api/pull/899) last year that introduced a regexp.

As I don't know much of the internals of Directus, I'd be happy to have a feedback on this, I can have easily missed something obvious.

Thanks you!

cc: @herrfinke

### DIFF
--- a/src/core/Directus/Database/TableGateway/BaseTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/BaseTableGateway.php
@@ -645,7 +645,24 @@ class BaseTableGateway extends TableGateway
     public function castFloatIfNumeric(&$value, $key)
     {
         if ($key != 'table_name') {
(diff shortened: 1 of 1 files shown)


apiraino:
@rijkvanzanten gently nudge for feedback :-)

Issue #1909, opened 2020-05-26
Create hooks not working with tables without auto generated primary keys


technicallyerik:
I'm attempting to create a `item.create` hook against a table that doesn't have auto generated primary keys.

Directus 8.6.2.

```
return [
    'actions' => [
        'item.create.widgets' => function ($data) {
                ...
        },
...
```

Unfortunately the `$data` parameter is coming back null.  

Digging around the code I believe it's because in `BaseTableGateway` it calls `getLastInsertValue` to get the inserted it and then refetches the data:

```
        if ($this->getTable() === SchemaManager::COLLECTION_COLLECTIONS) {
            $generatedValue = ArrayUtils::get($insertDataAssoc, $this->primaryKeyFieldName, 'table_name');
        } else {
            $generatedValue = $this->getLastInsertValue();
        }

        $resultData = $insertTableGateway->find($generatedValue);
```

I'm not entirely sure this is a correct enough solution to put in a PR, but this solution appears to be working for me locally:

```
        $generatedValue = $this->getLastInsertValue();
        if($generatedValue == 0) {
            $generatedValue = ArrayUtils::get($insertDataAssoc, $this->primaryKeyFieldName, 'table_name');
        }
```

apiraino:
@technicallyerik I've fell in this trap, too.

I have a `item.create.<collName>:after` hook which receives no payload. After a looong debugging session I've realized what the issue is: the record is not retrieved because the PK is an arbitrary field.

I'm still digging into Directus guts but I suspect the culprit could be also here:
https://github.com/directus/api/blob/master/src/core/Directus/Database/TableGateway/BaseTableGateway.php#L125

The table gateway is instantiated without providing PK field.

This issue is currently a blocker for us :-(

apiraino:
A possible side effect is another error I get after the record is created (adding more keywords to the issue):

```
{
  "error": {
    "code": null,
    "message": "Call to a member function toArray() on null",
    "class": "Error",
    "file": "/var/directus/src/core/Directus/Database/TableGateway/RelationalTableGateway.php",
    "line": 419
  }
}
```

```
<br />
<b>Notice</b>:  Undefined index: container_uid in <b>/var/directus/src/core/Directus/Database/TableGateway/BaseTableGateway.php</b> on line <b>414
</b><br />
```

(`container_uid` is my custom primaryKey field for the table)

technicallyerik:
@apiraino This doesn't appear to be fixed in 8.8.1 from the PR above.

[...]

PR #1939, opened 2020-06-20
Fix record creation when table PK is not the AUTO_INCREMENT field

apiraino:
Hello,

This is a draft to discuss a problem we have with hooks.

We have a table with such (very simplified) structure:

```
CREATE TABLE test_table (
  id int(11) NOT NULL AUTO_INCREMENT,
  pk_field bigint(11) NOT NULL PRIMARY KEY,
  name varchar(128)
);
```

This is our workflow:

- `test_table` is managed. The Directus App allows the user to only add a new record using the `name` field, the other two are hidden.
- After the form is submitted we have a `item.create.test_table:before` hook that populates the `pk_field`, the hook obviously only receives the `name` field.
- After the record is created we have another `item.create.test_table:after` hook that does other stuff and to do that needs the full record created before.

Now, for reasons that are not completely clear to me, the record fails to be created when the auto increment field does not match with the primary key field (Directus seems to always trust that they match).

I've dug into Directus internals but I'm not confident I got it right, but here's a tentative patch to start a discussion on this behaviour (which breaks our workflow).

I didn't investigate deeply:
- If this patch breaks other workflows (I don't have enough context)
- what's the performance cost of this patch (don't know enough of Zend framework)

By the way, I think this problem is also related to [this issue](https://github.com/directus/api/issues/1909).

I'd be really happy to receive feedback, comment and improvements on this one.

Thanks!

cc: @herrfinke

### DIFF
--- a/src/core/Directus/Database/TableGateway/BaseTableGateway.php
+++ b/src/core/Directus/Database/TableGateway/BaseTableGateway.php
@@ -398,9 +398,26 @@ class BaseTableGateway extends TableGateway
         if ($columnObject->hasAutoIncrement()) {
             $recordData[$primaryKey] = $TableGateway->getLastInsertValue();
         }
+        else if ($columnObject->hasPrimaryKey())
+        {
+            // identify the autoincrement field+value to retrieve the new record
+            $fieldAutoIncrement = null;
+            foreach($this->getTableSchema()->getFields() as $f) {
+                if ($f->hasAutoIncrement()) {
+                    $fieldAutoIncrement = $f->getName();
+                    $lastInsertValue = $TableGateway->getLastInsertValue();
+                    break;
+                }
+            }
+
+            // retrieve the full record
+            if (isset($fieldAutoIncrement)) {
+                $record = $this->findOneBy($fieldAutoIncrement, $lastInsertValue);
+                $recordData[$primaryKey] = $record[$primaryKey];
+            }
+        }
 
         $columns = SchemaService::getAllNonAliasCollectionFieldNames($this->table);
-
         return $TableGateway->fetchAll(function (Select $select) use ($recordData, $columns, $primaryKey) {
             $select
                 ->columns($columns)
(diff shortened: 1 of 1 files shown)


src/core/Directus/Database/TableGateway/BaseTableGateway.php
```diff
@@ -398,9 +398,26 @@ public function addRecordByArray(array $recordData)
...
+                    $fieldAutoIncrement = $f->getName();
+                    $lastInsertValue = $TableGateway->getLastInsertValue();
+                    break;
+                }
+            }
 
-        $columns = SchemaService::getAllNonAliasCollectionFieldNames($this->table);
+            // retrieve the full record
+            if (isset($fieldAutoIncrement)) {
+                $record = $this->findOneBy($fieldAutoIncrement, $lastInsertValue);
```
apiraino, line 415: Here I execute another query to retrieve the newly created record since the previous branch (`if $columnObject->hasAutoIncrement()`) can fail if `$primaryKey` is not also auto_increment

apiraino:
@rijkvanzanten gently nudge for feedback :-)

rijkvanzanten:
Does `getLastInsertValue` return just a primary key or the whole record?

apiraino:
> Does `getLastInsertValue` return just a primary key or the whole record?

The MySQL/MariaDB driver maps `getLastInsertValue` to [lastInsertValue](https://github.com/laminas/laminas-db/blob/5cfae8542111c48cfe1a9abd9fce948b00c9f3ee/src/TableGateway/AbstractTableGateway.php#L479) which executes [getLastGeneratedValue](https://github.com/laminas/laminas-db/blob/5cfae8542111c48cfe1a9abd9fce948b00c9f3ee/src/TableGateway/AbstractTableGateway.php#L309) which in the end returns the [last created id](https://github.com/laminas/laminas-db/blob/5cfae8542111c48cfe1a9abd9fce948b00c9f3ee/src/Adapter/Driver/Mysqli/Connection.php#L280).

I believe that historically MySQL/MariaDB does not support the `RETURNING` clause, therefore on INSERTs you can get only the PK created, while (for example) on Postgres you can [get the full record](https://github.com/laminas/laminas-db/blob/5cfae8542111c48cfe1a9abd9fce948b00c9f3ee/src/Adapter/Driver/Pgsql/Connection.php#L268) (or just some fields of it).

rijkvanzanten:
The `getLastGeneratedValue` part tripped me up a bit, I thought that was the generated auto increment value instead of the primary key 🙂

apiraino:
ah right, probably I've also misread your question :smile: 

thank you for merging!

src/core/Directus/Database/TableGateway/BaseTableGateway.php
```diff
@@ -798,6 +815,22 @@ protected function executeInsert(Insert $insert)
...
         $resultData = $insertTableGateway->find($generatedValue);
 
+        // try to retrieve the record using another field
+        if (empty($resultData)) {
+            $columnObject = $this->getTableSchema()->getField($this->primaryKeyFieldName);
+            if ($columnObject->hasPrimaryKey()) {
+                $fieldAutoIncrement = null;
+                foreach($this->getTableSchema()->getFields() as $f) {
+                    $fieldAutoIncrement = $f->getName();
+                    $lastInsertValue = $this->getLastInsertValue();
```
technicallyerik, line 825: This block of code doesn't really add anything, because if there is no auto generated id, `$this->getLastInsertValue()` return `0` and the `findOneBy` fails below.

PR #1944, opened 2020-06-25
Fix Google OAuth invalid_grant error

rdsubhas:
We're continuously seeing `invalid_grant` errors when trying to login with Google OAuth. This seems to be a very popular issue indeed: https://blog.timekit.io/google-oauth-invalid-grant-nightmare-and-how-to-fix-it-9f4efaf1da35

Testing some potential fixes. Will open PR when one of those work.


rdsubhas:
Sorry folks, none of the two settings work yet, am still trying to find out why this happens...

PR #1956, opened 2020-07-06
Corrects level of 'env' key/value for default/unknown project app configuration to prevent leaking of technical information

SoftwareEngineerChris:
I noticed that the ping app and "unknown" project app configurations always fell back to a `development` configuration. So when hitting an unknown endpoint (e.g. example.com/test) which would be routed via the unknown-project app, the response would contain details that should probably be kept private, such as PHP file paths and such. This is since these apps still use the regular error handlers, which return more information when in `development` configuration.

It looks like it _was_ the intention for these two apps to default to a `production` configuration, but because the `env` had been placed in an object called `app` instead of the root, the error handling code couldn't find it where it expected it, and therefore defaulted to "development".

As a side note, it looks like there are a few places in the codebase that default to `development` e.g. `->get('env', 'development')`. Perhaps these should default to `production` to be on the safer side?

Before fix:
Before

After fix:
After

### DIFF
--- a/src/helpers/app.php
+++ b/src/helpers/app.php
@@ -221,9 +221,7 @@ if (!function_exists('create_ping_server')) {
     function create_ping_server($basePath, array $config = [], array $values = [])
     {
         $app = create_app($basePath, array_merge([
-            'app' => [
-                'env' => 'production',
-            ],
+            'env' => 'production'
         ], $config), $values);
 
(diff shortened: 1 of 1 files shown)


edvasqueza:
will this patch be applied? I thinks it is really important

rijkvanzanten (approved):

Issue #1999, opened 2020-11-04
Uploading images with invalid chars in exif metadata fails


herrfinke:
### Problem
We have some images from different archives. Some of those could not be uploaded to directus.

### Detailed Description
It is not possible to upload images with invalid characters in the exif metadata description. The server just responded with a statuscode 500.
Those images were from different archives, and of course one could strip the exif metadata before uploading, but this is not helpful for endusers.

### Solution
Remove all non-utf8 characters from those exif fields before using them as default-values for the directus database.

I can create a PR with this solution.

### Stacktrace
```
[2020-11-03 16:59:17] api[tests].ERROR: Directus\Database\Exception\InvalidQueryException: Failed generating the SQL query. in /var/directus/src/core/Directus/Database/TableGateway/BaseTableGateway.php:818
Stack trace:
#0 /var/directus/vendor/zendframework/zend-db/src/TableGateway/AbstractTableGateway.php(264): Directus\Database\TableGateway\BaseTableGateway->executeInsert
#1 /var/directus/src/core/Directus/Database/TableGateway/BaseTableGateway.php(391): Zend\Db\TableGateway\AbstractTableGateway->insert
#2 /var/directus/src/core/Directus/Database/TableGateway/RelationalTableGateway.php(413): Directus\Database\TableGateway\BaseTableGateway->addRecordByArray
#3 /var/directus/src/core/Directus/Services/FilesServices.php(62): Directus\Database\TableGateway\RelationalTableGateway->createRecord
#4 /var/directus/src/endpoints/Files.php(74): Directus\Services\FilesServices->create
#5 [internal function]: Directus\Api\Routes\Files->create
#6 /var/directus/vendor/slim/slim/Slim/Handlers/Strategies/RequestResponse.php(40): call_user_func
#7 /var/directus/vendor/slim/slim/Slim/Route.php(281): Slim\Handlers\Strategies\RequestResponse->__invoke
#8 /var/directus/src/core/Directus/Application/Http/Middleware/AbstractRateLimitMiddleware.php(34): Slim\Route->__invoke
#9 [internal function]: Directus\Application\Http\Middleware\AbstractRateLimitMiddleware->__invoke
#10 /var/directus/vendor/slim/slim/Slim/DeferredCallable.php(57): call_user_func_array
#11 [internal function]: Slim\DeferredCallable->__invoke
#12 /var/directus/vendor/slim/slim/Slim/MiddlewareAwareTrait.php(70): call_user_func
[...]

PR #2001, opened 2020-11-05
ingnore invalid characters from exif-metadata

herrfinke:
solves #1999 

simply loop over all metadata fields which are used for metadata retrieval. Remove all non utf8 characters. 

Tested with different images, works fine and uploading of files with invalid exif-data is not resulting in sql error anymore.


rijkvanzanten (approved):

apiraino:
thanks @rijkvanzanten :tada:

PR #2391, opened 2021-02-03
Updating repository links to v8-archive

Ehryk:
This update is necessary to install directus now that the repository has been moved.


wytzejan:
@Ehryk Thanks for this pull request! I'm wondering how and when this pull request can be merged? I'm trying to install Directus v8 and having the same problems.

PR #2402, opened 2021-03-15
Fix/673

Rouzz:
Hi @philleepflorence ,

As discussed on the issue, I've put the proposed change in this PR so that it can be added to the official archived repo.
This is a big performance fix that we also applied to our local copy.

Please let me know if this proves to be an issue to be merged.

Issue : https://github.com/directus/v8-archive/issues/673

Thanks,
Rouzz

### DIFF
--- a/api/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
+++ b/api/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
@@ -1016,6 +1016,15 @@ class RelationalTableGateway extends BaseTableGateway
     }
 
     /**
+     * @param $params - GET Parameters
+     * @return int
+     */
+    private function countFilterTotal($params)
+    {
+        return (int)$this->fetchItems($params, null, true)[0]['total'];
+    }
+
+    /**
      * Updates Metadata Object with Pagination
      *
      * @param $metadata - Existing metadata object
@@ -1040,23 +1049,11 @@ class RelationalTableGateway extends BaseTableGateway
 
         $meta_param = explode(',', ArrayUtils::get($params, 'meta', ''));
 
-        if ((in_array('filter_count', $meta_param) || in_array('*', $meta_param))) {
-            $metadata['filter_count'] = $countedData['total_count'];
+        //filter_count performance hotfix
+        if ($filtered && (in_array('filter_count', $meta_param) || in_array('*', $meta_param) || in_array('page', $meta_param))) {
+            $metadata['filter_count'] = $this->countFilterTotal($params);
         }
-
-        if ($filtered) {
-            $filteredparams = array_merge($params, [
-                "depth" => 0,
-                "fields" => $this->primaryKeyFieldName,
-                "limit" => -1
-            ]);
-
-            $entries = $this->fetchItems($filteredparams);
-            $total = count($entries);
-            if (in_array('filter_count', $meta_param) || in_array('*', $meta_param)) {
-                $metadata['filter_count'] = $total;
-            }
-        }
+        $total = $metadata['filter_count'] ?: $countedData['total_count'];
 
         $limit = $limit < 1 ? $rows : $limit;
         $pages = $total ? ceil($total / $limit) : 1;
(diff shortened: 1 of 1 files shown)


api/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
```diff
@@ -1040,23 +1049,10 @@ public function createMetadataPagination(array $metadata = [], array $params = [
...
-
-            $entries = $this->fetchItems($filteredparams);
-            $total = count($entries);
-            if (in_array('filter_count', $meta_param) || in_array('*', $meta_param)) {
-                $metadata['filter_count'] = $total;
-            }
+        if ($filtered && (in_array('filter_count', $meta_param) || in_array('*', $meta_param) || in_array('page', $meta_param))) {
+            $metadata['filter_count'] = $this->countFilterTotal($params);
         }
+        $total = $metadata['filter_count'] ?: $countedData['total_count'];
```
revan57, line 1055: this logic will not return 'filter_count' when you set an empty value, for example, to filter field in admin panel, i think, it should be like this:
```php
if ($filtered) {
            $total = $this->countFilterTotal($params);
            $filtered_count = $total;
        }
        if ((in_array('filter_count', $meta_param) || in_array('*', $meta_param))) {
            $metadata['filter_count'] = isset($filtered_count) ? $filtered_count : $countedData['total_count'];
        }
```

api/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
```diff
@@ -1138,6 +1133,7 @@ public function fetchItems(array $params = [], \Closure $queryCallback = null)
             }
         }
 
+        $selectedFields = $countOnly ? ['total' => new Expression('COUNT(*)')] : $selectedFields;
```
revan57, line 1136: when pagination is used in request, there will be an 'offset' & 'limit' params, so, sql query like 'select count(*) from table limit n offset x' (x > 0) will return empty set, so you should unset it like:

```php
if ($countOnly) {
            $selectedFields = ['total' => new Expression('COUNT(*)')];
            unset($params['offset']);
            unset($params['limit']);
            unset($params['sort']);
        }
```

api/src/core/Directus/Database/TableGateway/RelationalTableGateway.php
```diff
@@ -1040,23 +1049,10 @@ public function createMetadataPagination(array $metadata = [], array $params = [
...
-                "fields" => $this->primaryKeyFieldName,
-                "limit" => -1
-            ]);
-
-            $entries = $this->fetchItems($filteredparams);
-            $total = count($entries);
-            if (in_array('filter_count', $meta_param) || in_array('*', $meta_param)) {
-                $metadata['filter_count'] = $total;
-            }
+        if ($filtered && (in_array('filter_count', $meta_param) || in_array('*', $meta_param) || in_array('page', $meta_param))) {
```
revan57, line 1052: admin panel uses 'filter_count' param to display num of elements in collection, in your case, if there will be no filters, but 'filter_count' is defined in request query, it won't return it.
The second moment is that $total variable, declared above, becomes useless. I think it should be like this:

```php
if ($filtered) {
            $total = $this->countFilterTotal($params);
            $filtered_count = $total;
}
if ((in_array('filter_count', $meta_param) || in_array('*', $meta_param))) {
        $metadata['filter_count'] = isset($filtered_count) ? $filtered_count : $countedData['total_count'];
}
```

westhouseit:
@Rouzz any thoughts on adjusting your pull request with the suggestions by @revan57 above?

Rouzz:
@westhouseit Gonna be honest, this is just copy pasting from the comments of the original issues as it was fixing the issue we were having and the rest of the team wasn't familiar with github. But I'm not familiar enough with the issue or the product itself to have any opinion on @revan57 's propositions. Don't even have a project to test it on.

I may try to get to it as some point, but very much doubt it will be soon. If someone wants to take this over that's very fine by me

